Issue #11, opened 2016-10-03
PROCESSING_ERROR should be properly showed on main view


Skejven:
On the left hand side and test level it seems fail (see the screenshots).
But on the url level it seems success.

While checking metadata it looks like the status is PROCESSING_ERROR however the message is not displayed.

There are other collector statuses that should be supported. See:
- `com.cognifide.aet.communication.api.metadata.CollectorStepResult.Status`

The colors should be consistent on all levels; the message should be displayed.

_Note:_
`DUPLICATES_PATTERN` means that collector doesn't store new version in database (as it is same as pattern),

| Task | Status |
| --- | --- |
| if there is processing error it should be displayed in test case report |  |
| message (description) is provided for error (copied from metadata) |  |

![test-view](https://cloud.githubusercontent.com/assets/19705838/19032916/a9edf022-895c-11e6-8ae5-4e68da962c3d.png)
![url-view](https://cloud.githubusercontent.com/assets/19705838/19032917/a9ee7f6a-895c-11e6-86ce-89e660d1d479.png)
![metadata](https://cloud.githubusercontent.com/assets/19705838/19032918/a9f0caa4-895c-11e6-868e-4c4992bd8148.png)

PR #17, opened 2016-10-12
Feature/#11 display processing errors

Skejven:
This code is for #11 - processing errors are now displayed on report tabs. 

---


report/src/main/webapp/app/services/metadataEndpoint.service.js
```diff
@@ -15,20 +15,20 @@
...
 	 */
-	function MetadataEndpointService($q, $http, configService, requestParametersService) {
+	function MetadataEndpointService($q, $http, endpointConfiguration, requestParametersService) {
 		var service = {
 				getMetadata: getMetadata,
 				saveMetadata: saveMetadata
 			},
 			requestParams = requestParametersService.get(),
-			configParams = configService.getConfig();
+			getEndpointUrl = endpointConfiguration.getEndpointUrl();
```
wiiitek: Could we rename `getEndpointUrl` variable? To sth like
- `metadataUrl`
- `endpoint`
- ??

report/src/main/webapp/app/services/endpointConfiguration.service.js
```diff
@@ -16,17 +16,16 @@
...
-				};
-
-				return config;
-			}
-		};
-	});
+  'use strict';
+  angularAMD.factory('endpointConfiguration', function () {
+    return {
+      getEndpointUrl: function () {
```
wiiitek: As service is called `endpointConfiguration~maybe method could be renmd to `getUrl` ?
What do you think?

Skejven: Yes, this is good idea - I renamed method to `getEndpoint` and `production` to `getUrl`.

wiiitek:
For the source comparator there is a JavaScript error as Angular expects JSON object but artifactId is a HTML source. This should probably be investigated...

PR #18, opened 2016-10-15
fix for client side performance report

wiiitek:
Some fixes for client-side-performance reports.
Refactoring / code formatting could be done later.


wiiitek:
___

PR #22, opened 2016-10-18
Search filter for test renamed

wiiitek:
___


report/src/main/webapp/app/layout/main/test/mainView.test.view.html
```diff
@@ -20,7 +20,7 @@
 <div class="main page-main">
     <div class="row reports-list">
         <div class="reports-list-item {{url.getStatus()}}"
-             ng-repeat="url in testView.urls | aetTestUrlsStatusFilter:$root.activeFilters | aetTestUrlsSearchFilter:$root.searchText | orderBy:'name'"
+             ng-repeat="url in testView.urls | aetTestUrlsStatusFilter:$root.activeFilters | aetTestSearchFilter:$root.searchText | orderBy:'name'"
```
Skejven: We probably should remove usage of this filter from this level. It looks like it causes a bug, than when search is applied, no url is visible on Test View level (see implementation that takes `tests` array).

PR #24, opened 2016-10-20
Css for url tabs

wiiitek:
___


wiiitek:
### Before

[](https://postimg.org/image/mp1jjogo1/)

### After

[](https://postimg.org/image/6i3w6ax0n/)

PR #25, opened 2016-10-22
fix for old cookbook attribute

wiiitek:
Atribute for Karaf debug should be `enable_debug`. See [aet-cookbook atributes](https://github.com/Cognifide/aet-cookbook#attributes) and [aet-cookbook source code](https://github.com/Cognifide/aet-cookbook/blob/master/attributes/karaf.rb#L36).


wiiitek:
___

PR #26, opened 2016-10-22
Fix for color names and refactoring

wiiitek:
Unit test introduced for private method: `ImageComparison#convertImageTo2DArray`.
This could be controversial, but the unit test runs this important method in separation.
Allows to validate/understand how it works.

---


core/jobs/src/main/java/com/cognifide/aet/job/common/comparators/layout/utils/ImageComparison.java
```diff
@@ -124,22 +120,23 @@ private static int fastCompareMatchingArea(int minWidth, int minHeight, Writable
...
+  // package-scoped for unit test
+  static int[][] convertImageTo2DArray(BufferedImage image) {
+    final byte[] imageBytes = ((DataBufferByte) image.getRaster().getDataBuffer()).getData();
     final int width = image.getWidth();
     final int height = image.getHeight();
 
     int[][] result = new int[height][width];
-    final int pixelLength = 4;
     int row = 0;
     int col = 0;
```
wiiitek: Lines below are the key point of this pull request.

PR #27, opened 2016-10-25
fix for tabs buttons being scrolled away

wiiitek:
---


wiiitek:
This is a fix for test case tabs being scrolled away for long results
See video for the bug at:
https://sendvid.com/ru5bj2bv

[](https://s13.postimg.org/k7f1ijw2v/scroll_bug.png)

PR #28, opened 2016-11-02
Feature/#12 stabilization of login modifier

tomaszhulakcogni:


core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/login/LoginModifierConfig.java
```diff
@@ -0,0 +1,148 @@
...
+
+import com.cognifide.aet.job.api.ParametersValidator;
+import com.cognifide.aet.job.api.exceptions.ParametersException;
+import org.apache.commons.lang3.BooleanUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.math.NumberUtils;
+
+import java.util.Map;
+
+public class LoginModifierConfig {
```
wiiitek: Could we have this class package-scoped?

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/login/LoginModifier.java
```diff
@@ -135,35 +78,46 @@ public CollectorStepResult collect() throws ProcessingException {
...
+
+  private void loginToForm() throws ProcessingException {
+    LOGGER.info("Attempting to login user: {} on site {}", config.getLogin(), config.getLoginPage());
+    webDriver.get(config.getLoginPage());
+    LoginFormComponent form = new LoginFormComponent(webDriver, config.getLoginInputSelector(),
+            config.getPasswordInputSelector(), config.getSubmitButtonSelector());
+    form.login(config.getLogin(), config.getPassword());
+  }
+
+  private void sleep() {
```
wiiitek: Could we rename this method to sth like:
- `waitForAuthentication`
- `delayBeforeLoginCheck`
- ??

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/login/LoginModifierConfig.java
```diff
@@ -0,0 +1,148 @@
...
+    this.loginPage = params.get(LOGIN_PAGE_PARAM);
+    this.login = getParameter(params, LOGIN_PARAM, DEFAULT_LOGIN);
+    this.password = getParameter(params, PASSWORD_PARAM, DEFAULT_PASSWORD);
+    this.loginInputSelector = getParameter(params, LOGIN_INPUT_SELECTOR_PARAM, DEFAULT_LOGIN_INPUT_SELECTOR);
+    this.passwordInputSelector = getParameter(params, PASSWORD_INPUT_SELECTOR_PARAM, DEFAULT_PASSWORD_INPUT_SELECTOR);
+    this.submitButtonSelector = getParameter(params, SUBMIT_BUTTON_SELECTOR_PARAM, DEFAULT_SUBMIT_BUTTON_SELECTOR);
+    this.loginTokenKey = getParameter(params, LOGIN_TOKEN_KEY_PARAM, DEFAULT_LOGIN_TOKEN);
+    this.forceLogin = BooleanUtils.toBoolean(params.get(FORCE_LOGIN));
+    int loginTimeout = NumberUtils.toInt(getParameter(params, LOGIN_CHECK_TIMEOUT_PARAM, DEFAULT_CHECK_TIMEOUT_PARAM));
+    this.loginCheckTimeout = Math.min(10000, loginTimeout);
```
wiiitek: Could we rename this attribute to sth like:
- `this.waitForAuthentication = loginCheckDelay`
- `this.delayBeforeLoginCheck = delay`
- ??

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/login/LoginModifier.java
```diff
@@ -135,35 +78,46 @@ public CollectorStepResult collect() throws ProcessingException {
...
+    if (config.getLoginCheckTimeout() > 0) {
+      LOGGER.info("Waiting {} ms.", config.getLoginCheckTimeout());
+      try {
+        Thread.sleep(config.getLoginCheckTimeout());
+      } catch (InterruptedException e) {
+        LOGGER.error("Interruption", e);
+        Thread.currentThread().interrupt();
```
wiiitek: Hmm. This line looks suspicious.
I went through one article about the topic:
    http://www.javaspecialists.eu/archive/Issue056.html
and it looks like it makes sense for interrupting newly created threads.
Here however we are in some thread that is performing screen collection.
I woud rather throw some runtime exception here:
`new RuntimeException(this thread should NOT be interrupted here)`..

Could we consult someone on this?

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/login/LoginModifier.java
```diff
@@ -36,63 +32,17 @@
...
-
-  private String loginTokenKey;
-
-  private int loginCheckTimeout;
-
-  private boolean forceLogin;
+  private LoginModifierConfig config;
 
   LoginModifier(WebCommunicationWrapper webCommunicationWrapper, ValidationResultBuilder validationResultBuilder) {
     this.webCommunicationWrapper = webCommunicationWrapper;
```
wiiitek: Could we also have `this.webDriver` in line 49 ?

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/login/LoginModifier.java
```diff
@@ -135,35 +78,46 @@ public CollectorStepResult collect() throws ProcessingException {
...
+  private void loginToForm() throws ProcessingException {
+    LOGGER.info("Attempting to login user: {} on site {}", config.getLogin(), config.getLoginPage());
+    webDriver.get(config.getLoginPage());
+    LoginFormComponent form = new LoginFormComponent(webDriver, config.getLoginInputSelector(),
+            config.getPasswordInputSelector(), config.getSubmitButtonSelector());
+    form.login(config.getLogin(), config.getPassword());
+  }
+
+  private void sleep() {
+    if (config.getLoginCheckTimeout() > 0) {
```
Skejven: We use `sleep` in two places. Its action depends on `LoginCheckTimeout` value. Should `sleep` method be used both before each re-attempt of login and during `login` method?

integration-tests/test-suite/pom.xml
```diff
@@ -45,7 +45,7 @@
 
   <profiles>
     <profile>
-      <!-- Assembly test suites from 'partials' catolog into one main-test.xml -->
+      <!-- Assembly test suites from 'partials' catalog into one main-test.xml -->
```
mateuszgrab-cognifide: Please avoid changes with only white-space characters modification.

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/login/LoginModifier.java
Skejven: I think this is so little damage here when `sleep` is interrupted, we can go with this approach. It looks that `SleepModifier` handles interruption in similar way - it may be area to improve at least logging there.

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/login/LoginModifier.java
tomaszhulakcogni: Yes. I'm going to describe it in the documentation.

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/login/LoginModifier.java
tomaszhulakcogni: Renamed to `delayBeforeLoginCheckOrReattempt`. Long, but descriptive - this timeout is used in two contexts.

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/login/LoginModifierConfig.java
tomaszhulakcogni: Done

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/login/LoginModifier.java
tomaszhulakcogni: Done

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/login/LoginModifier.java
tomaszhulakcogni: Left as it is.

Skejven:
Please update [Login Modifier Documentation|https://github.com/Cognifide/aet/blob/master/documentation/src/main/wiki/LoginModifier.md] in this pull request.

PR #29, opened 2016-11-06
fix for ignored test

wiiitek:
___


core/runner/src/test/java/com/cognifide/aet/runner/distribution/dispatch/CollectorJobSchedulerTest.java
```diff
@@ -95,17 +95,19 @@ public void add_expectMessageSent() throws Exception {
...
 
-    Queue<MessageWithDestination> messagesQueue1 = mockMessagesQueue(Mockito.mock(Message.class), Mockito.mock
-            (Destination.class), correlationID);
-    Queue<MessageWithDestination> messagesQueue2 = mockMessagesQueue(Mockito.mock(Message.class), Mockito.mock
-            (Destination.class), correlationID);
     tested.add(messagesQueue1, correlationID);
+    // messageQueues will be processed and removed from messagesMap
+    // in safeRun method of CollectorJobScheduler
+    // so it is possible to add another with the same correlationId
+    Thread.sleep(1000);
```
Skejven: Would this be possible to make here some active waiting? So that when we run this test on a slow machine this test will not fail (now it may happen that this 1 second is not enough).
In other way pleas remove this test.

core/runner/src/test/java/com/cognifide/aet/runner/distribution/dispatch/CollectorJobSchedulerTest.java
```diff
@@ -95,17 +95,19 @@ public void add_expectMessageSent() throws Exception {
...
     tested.add(messagesQueue2, correlationID);
```
toniedzwiedz: More importantly, what does this method actually test? I may be missing something but I fail to see an assertion of any sort.

core/runner/src/test/java/com/cognifide/aet/runner/distribution/dispatch/CollectorJobSchedulerTest.java
wiiitek: Hi @Skejven , I had a closer look into the test and the class and need to consult something:

the flow that i debugged on Windows with IntelliJ is as follows:

1. wait for `availableMessages` semaphore
2. *messege queue 1* is added to `messagesMap`
3. *message queue 1* is processed in `safeRun` method and sent
4. wait for `availableMessages` semaphore
5. `safeRun` method checks that the *message queue 1* has no more entries
6. *message queue 1* is removed from `messagesMap`

There may be a race condition. If the `safeRun` method is quick enough to make two loops, then it consumes the *message queue 1* and also removes it from `messageMap`. Otherwise (if only one loop happens) there will be exception thrown while adding second message queue with same correlation ID.

As this issue is complicated, could we have someone to review this?
I have set breakpoints at the two lines where the messege queues are added and also in `CollectorJobSchedule`: lines 255, 129.

I will add another pull request with proposed solution...

wiiitek:
I have updated the pull request.
@Skejven , @toniedzwiedz , what do you think about it?

core/runner/src/test/java/com/cognifide/aet/runner/distribution/dispatch/CollectorJobSchedulerTest.java
```diff
@@ -96,15 +96,19 @@ public void add_expectMessageSent() throws Exception {
   }
 
 
-  @Ignore("Ignored for some unexpected test failure on Linux")
-  @Test(timeout = 2000l, expected = IllegalStateException.class)
-  public void add_whenAddedMessageExists_expectIllegalStateException() throws Exception {
+  @Test(expected = IllegalStateException.class)
```
toniedzwiedz: Now I can see that the thing we assert here is that an exception is thrown. However, it's still unclear to me which method all is expected to throw an exception and why. 

There are better ways to unit-test exceptions than specifying the `expected` parameter of the `@Test` annotation.

You can be a bit more expressive if you use the [`ExpectedException` rule](https://github.com/junit-team/junit4/wiki/exception-testing#expectedexception-rule). 

Personally, I find this library useful when testing exceptions: https://github.com/Codearte/catch-exception

wiiitek: Thank you @toniedzwiedz ! `@Rule` approach applied. I will take a look at the suggested library soon.

wiiitek:
Tested and ready to be merged.

core/runner/src/test/java/com/cognifide/aet/runner/distribution/dispatch/CollectorJobSchedulerTest.java
toniedzwiedz: As an alternative to the library, you could consider upgrading to JUnit 5 which has some neat OOTB support for exception testing

    @Test
    void exceptionTesting() {
        Throwable exception = assertThrows(IllegalArgumentException.class, () -> {
            tested.someMethod(illegalArgument);
        });
        assertThat(exception.getMessage(), is("Some expected message"));
    }

PR #30, opened 2016-11-17
fixed cookie action test report display

Skejven:
Fixed cookie report display in 'test' action mode.
___

### DIFF
--- a/report/src/main/webapp/app/layout/main/url/reports/cookie.html
+++ b/report/src/main/webapp/app/layout/main/url/reports/cookie.html
@@ -69,16 +69,11 @@
     </table>
   </div>
 
-  <div ng-if="case.comparator.parameters.action == 'test'">
-    <div ng-if="case.comparator.stepResult.status == 'FAILED'">
-      Cookie with name: <span class="status failed">{{case.result.cookieName}}</span>
-      <span ng-if="case.result.cookieValue"> and value:
-      <span class="status failed">{{case.result.cookieValue}}</span></span> not found
-    </div>
-    <div ng-if="case.comparator.stepResult.status != 'FAILED'">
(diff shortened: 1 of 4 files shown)


report/src/main/webapp/app/layout/main/url/reports/cookie.html
```diff
@@ -67,20 +67,18 @@
...
+  <div ng-if="case.comparator.parameters.action == 'test'">
+    <div ng-if="case.comparator.stepResult.status == 'FAILED'">
+      Cookie with name: <span class="status failed">{{case.result.cookieName}}</span>
+      <span ng-if="case.result.cookieValue"> and value:
+      <span class="status failed">{{case.result.cookieValue}}</span></span> not found
+    </div>
+    <div ng-if="case.comparator.stepResult.status != 'FAILED'">
+      Cookie with name: <span class="status passed">{{case.result.cookieName}}</span> <span ng-if="case.result.cookieValue">
+      and value: <span class="status passed">{{case.result.cookieValue}}</span></span> found
+    </div>
```
tomaszhulakcogni: Is it possible to avoid duplication in both cases: when it failed and when it succeeded? The only difference is `failed`/`passed` class and the word "not".

How about:

```
<div ng-if="case.comparator.parameters.action == 'test'" ng-class="case.comparator.stepResult.status == 'FAILED' ? 'failed' : 'passed'">
    Cookie with name: <span class="status">{{case.result.cookieName}}</span> 
    <span ng-if="case.result.cookieValue"> and value: <span class="status passed">{{case.result.cookieValue}}</span></span>
    <span ng-if="case.comparator.stepResult.status == 'FAILED'">not</span> found
</div>
```

PR #44, opened 2017-01-13
Screenshots of individual elements of the page

subiron:
added ability to take screenshots of parts of the page by using css or xpath element selectors

Please do not merge yet! 
(Base branch is set to "modifiers_enhancements" for clear diff until modifiers_enhancements be merged into master branch)
___


core/jobs/src/main/java/com/cognifide/aet/job/common/collectors/screen/ScreenCollector.java
```diff
@@ -22,76 +22,109 @@
 import com.cognifide.aet.job.api.collector.CollectorProperties;
```
subiron: done

core/jobs/src/main/java/com/cognifide/aet/job/common/collectors/screen/ScreenCollector.java
```diff
@@ -83,15 +97,62 @@ private boolean isPatternAndResultMD5Identical(byte[] screenshot) {
 
   @Override
   public void setParameters(Map<String, String> params) throws ParametersException {
-    //No parameters here
+    if (StringUtils.isNotBlank(params.get(XPATH_PARAM)) || StringUtils
```
Skejven: Maybe we could move setting parameters from `WebElementsLocatorParams` to some method in `WebElementsLocatorParams`?

core/jobs/src/main/java/com/cognifide/aet/job/common/collectors/screen/ScreenCollector.java
```diff
@@ -83,15 +97,62 @@ private boolean isPatternAndResultMD5Identical(byte[] screenshot) {
...
+    } catch (IOException e) {
+      throw new ProcessingException("Unable to create image from taken screenshot", e);
+    } finally {
+      IOUtils.closeQuietly(in);
     }
   }
 
+  private byte[] bufferedImageToByteArray(BufferedImage bufferedImage) throws ProcessingException {
+    ByteArrayOutputStream temporaryStream = new ByteArrayOutputStream();
+    try {
```
Skejven: You may get benefit from `Closable` interface that ByteArrayOutputStream implements:
```java
    try (ByteArrayOutputStream temporaryStream = new ByteArrayOutputStream()){
      ImageIO.write(bufferedImage, PNG_FORMAT, temporaryStream);
      temporaryStream.flush();
      return temporaryStream.toByteArray();
    } catch (IOException e) {
      throw new ProcessingException("Unable to convert screenshot part to byte Array", e);
    }
```

documentation/src/main/wiki/ScreenCollector.md
```diff
@@ -6,14 +6,22 @@
...
 
 **Note that you cannot maximize the window and specify the dimension at the same time. If no parameters provided, default browser size is set before taking screenshot.**
 
+##### Parameters
+
+| Parameter | Value | Description | Mandatory |
+| --------- | ----- | ----------- | --------- |
+| `xpath` | xpath_to_element | Xpath to element(s) | optional (either xpath or css) |
+| `css` | css_selector_to_element | css selector to element(s)| optional (either xpath or css) |
+| `timeout` | 1000ms | The timeout for the element to appear, in milliseconds. The max value of this parameter is 15000 milliseconds (15 seconds). | no (default will be used) |
```
Skejven: It looks that timeout parameter is used only when one of selectors is used. It should be mentioned here unless its possible to update the code (which would be much more awesome!).

integration-tests/test-suite/partials/layout.xml
```diff
@@ -118,5 +118,19 @@
 			<url href="comparators/layout/failed.jsp" />
 		</urls>
 	</test>
+
+	<test name="comparator-Layout-Partial-SUCCESS">
```
Skejven: Please provide one more test for `xpath` selector.

core/jobs/src/main/java/com/cognifide/aet/job/common/collectors/screen/ScreenCollector.java
subiron: it is, but this is extra check for screen collector as both of those params are optional.

core/jobs/src/main/java/com/cognifide/aet/job/common/collectors/screen/ScreenCollector.java
subiron: I always forget about that

documentation/src/main/wiki/ScreenCollector.md
subiron: you mean some kind if warning message when timeout param is provided but neither xpath or css ?

Skejven: I'd rather thought about mentioning at documentation, that `timeout` parameter is applicable only when taking a screenshot of a page element. It does not work in `fullScreen` mode.
But I would be much more happier if `timeout` would also work in `fullScreen` mode and there would be no need to change docs then.

subiron: > But I would be much more happier if timeout would also work in fullScreen mode and there would be no need to change docs then.

I have no idea how to (for what purpose) use timeout param when taking fullscreen-shot 
(we have sleep modifier for that)
imo just update doc, what do you think ?

PR #51, opened 2017-01-29
Improvements for modifiers

wiiitek:


core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/click/ClickModifier.java
```diff
@@ -50,19 +50,12 @@ public CollectorStepResult collect() throws ProcessingException {
...
-        result = CollectorStepResult.newProcessingErrorResult(message);
-        LOG.warn(message);
-      }
+      elementToClick.click();
+      result = CollectorStepResult.newModifierResult();
     } catch (WebDriverException e) {
       final String message =
-          String.format("No element defined by %s could be found before timeout (%s seconds)!",
-              getLocator().toString(), getTimeoutInSeconds());
+          String.format("Element not found before timeout (%s seconds): %s!",
```
mchrominski: I really dislike error messages with the exclamation mark in the end. 
Obviously such sentence should not be provided to the end user - it's part of the development activities. But the communication interface may (should?) stay the same.

I couldn't find any Java standards approach, but as stated here http://stackoverflow.com/a/34136055/774328 the C# documentation explicitely discourages that (https://msdn.microsoft.com/en-us/library/system.exception.message%28v=vs.110%29.aspx) 

> Do not use question marks or exclamation points

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/click/ClickModifier.java
```diff
@@ -50,19 +50,12 @@ public CollectorStepResult collect() throws ProcessingException {
...
+                  getTimeoutInSeconds(), getLocator().toString());
       LOG.warn(message, e.getMessage());
```
toniedzwiedz: Why log the message and not the throwable itself? Are we hiding the stack trace on purpose?

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/click/ClickModifier.java
subiron: "Element not found " may be misleading here, there are other reasons when 'WebDriverException' can be thrown
 e.g. when element 'X' exist (and can be found using locator) but it's 'unclickable', because another element 'Y' is drawn on top of element 'X'.

wiiitek: I will remove the `!` :)

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/click/ClickModifier.java
wiiitek: I will reorganize the code: the TimeoutException is happening quite often for elements not present on some pages.

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/click/ClickModifier.java
wiiitek: I will add another `catch` block for TimeoutException :)

wiiitek: I will remove exclamatin mark :)

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/click/ClickModifier.java
wiiitek: I will fix it. Thank you!

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/click/ClickModifier.java
wiiitek: You are right, I will fix it. Thank you!

PR #52, opened 2017-02-05
Update for documentation and default proxy

wiiitek:
As confirmed with @Skejven and @subiron the "embedded" proxy is no longer available in AET.


wiiitek:
___

PR #53, opened 2017-02-05
Removing not needed parameter

wiiitek:
As there is no embedded proxy we don't need a port configuration for IT.
REST proxy has its own configuration (with a port number).

### DIFF
--- a/osgi-dependencies/proxy/src/main/java/com/cognifide/aet/proxy/ProxyServerProvider.java
+++ b/osgi-dependencies/proxy/src/main/java/com/cognifide/aet/proxy/ProxyServerProvider.java
@@ -44,7 +44,7 @@ public class ProxyServerProvider {
           unbind = "unbindProxyManager")
   private final Map<String, ProxyManager> collectorManagers = Maps.newConcurrentMap();
 
-  public ProxyServerWrapper createProxy(String useProxy, int port) throws ProxyException {
+  public ProxyServerWrapper createProxy(String useProxy) throws ProxyException {
     String proxyType = useProxy;
     if ("true".equals(useProxy)) {
       proxyType = DEFAULT_PROXY_MANAGER;
@@ -54,9 +54,10 @@ public class ProxyServerProvider {
       throw new ProxyException("Undefined ProxyManager with proxyType: " + proxyType);
     }
     try {
-      return proxyManager.createProxy(port);
+      return proxyManager.createProxy();
     } catch (ProxyException e) {
-      throw new ProxyException("Unable to create ProxyServer on port " + port + " by ProxyManager", e);
+      String managerClass = proxyManager.getClass().getCanonicalName();
+      throw new ProxyException("Unable to create ProxyServer with ProxyManager: " + managerClass, e);
     }
   }
 
--- a/osgi-dependencies/proxy/src/main/java/com/cognifide/aet/proxy/ProxyServerProvider.java
+++ b/osgi-dependencies/proxy/src/main/java/com/cognifide/aet/proxy/ProxyServerProvider.java
@@ -44,12 +44,10 @@ public class ProxyServerProvider {
           unbind = "unbindProxyManager")
   private final Map<String, ProxyManager> collectorManagers = Maps.newConcurrentMap();
 
-  public ProxyServerWrapper createProxy(String proxyTypeConfig) throws ProxyException {
-    String proxyType;
-    if ("true".equals(proxyTypeConfig)) {
+  public ProxyServerWrapper createProxy(String useProxy) throws ProxyException {
+    String proxyType = useProxy;
+    if ("true".equals(useProxy)) {
       proxyType = DEFAULT_PROXY_MANAGER;
-    } else {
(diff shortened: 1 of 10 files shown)


osgi-dependencies/proxy/src/main/java/com/cognifide/aet/proxy/ProxyServerProvider.java
```diff
@@ -44,19 +44,20 @@
           unbind = "unbindProxyManager")
   private final Map<String, ProxyManager> collectorManagers = Maps.newConcurrentMap();
 
-  public ProxyServerWrapper createProxy(String useProxy, int port) throws ProxyException {
-    String proxyType = useProxy;
-    if ("true".equals(useProxy)) {
+  public ProxyServerWrapper createProxy(String usedProxy) throws ProxyException {
+    String proxyType = usedProxy;
+    if ("true".equals(usedProxy)) {
```
toniedzwiedz: String proxyType = usedProxy;
    if ("true".equals(usedProxy)) {
   //...
   proxyType = DEFAULT_PROXY_MANAGER;

The values that you assign to the `proxyType` variable here are `"true"` and `"embedded"`... That's stringly typed and difficult to follow.

Did you actually mean to use the same variable? If so, consider using an explicit `enum` type to cover the possible values. If not, use separate variables for these values and use `boolean` when actually handling boolean variables.

wiiitek: Do I understand correctly you don't like having `"true"` passed as sting type value?
I have renamed the variables in order to make the code more clear.
However:
- it may be bad idea to use **enum** for proxy types, as we don't know how many type there will be in an AET system. someone could add new `ProxyManager` service with new type.
- the `"true"` value is passed from suite XML configuration but the [`useProxy` attribute](https://github.com/Cognifide/aet/wiki/SuiteStructure#proxy) could have non-boolean values.

I can't make this code better :( Could you provide some code samples / commits / pull requests?

toniedzwiedz: There are two things I don't like

1. Passing boolean values as Strings. IMO, as soon as the value is read from the file, it should be converted to the most specific type possible in a given context.
2.  Mixing values with seemingly disparate semantics. 

Let's have a look at the current code.

    String proxyType;
    if ("true".equals(proxyTypeConfig)) {
        proxyType = DEFAULT_PROXY_MANAGER;
    } else {
        proxyType = proxyTypeConfig;
    }

It looks better now. `"true"` does not look like third alternative to `"embedded"` and `"rest`".

I'm not sure I understand the meaning of the name `proxyTypeConfig`. What exactly does it mean? That the default config should be used? If so, rename it to something along the lines of `useDefaultProxyType`. Can you show me an example of the XML file where this is configured?

On a side note, you could get rid of the null initialisation by converting the `if-else` block to a simple assignment:

  String proxyType = "true".equals(proxyTypeConfig) ? DEFAULT_PROXY_MANAGER : proxyTypeConfig;

Finally, if `proxyTypeConfig` is a boolean, the constructor should probably not take a `String` parameter.

wiiitek: Thank you @toniedzwiedz . I have removed the changes related to webProxy config.
They should be consulted and addressed in another pull request in future.

Skejven: This is clearly some configuration legacy. Before Browsermob, AET used another proxy (embedded) and then `useProxy` configuration could have 2 values: `true` or `false`. When the browsermob came, because of a backward compatibility, configuration was extended to `rest` and `embedded`.
In time, embedded proxy option was removed from the system as it didn't work well.
Thank you @wiiitek for noticing this.
I think we have 2 paths now:
- lets leave `useProxy` possible values as `true` and `rest` which are equivalent (`true` indicates that the *default* proxy will be used).
- there will be only named type of a proxy (remove `true` as a possible value) after deprecating it first for e.g. 2 minor versions?

What do you think?

toniedzwiedz: @Skejven 

> - lets leave useProxy possible values as `true` and `rest` which are equivalent (`true` indicates that the _default_ proxy will be used).

Why not `default` and `rest` then?

> - there will be only named type of a proxy (remove `true` as a possible value) after deprecating it first for e.g. 2 minor versions?

Sounds good too.

wiiitek:
___

osgi-dependencies/proxy/src/main/java/com/cognifide/aet/proxy/ProxyServerProvider.java
wiiitek: I have created new issue for that: https://github.com/Cognifide/aet/issues/56

Could you please approve / update the description when you have time?

PR #54, opened 2017-02-05
Upgrade for bobcat from 0.9.6 to 1.1.3

wiiitek:
___

### DIFF
--- a/integration-tests/sanity-functional/pom.xml
+++ b/integration-tests/sanity-functional/pom.xml
@@ -31,10 +31,12 @@
   <description>BDD tests for AET</description>
 
   <properties>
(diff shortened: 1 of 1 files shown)


integration-tests/sanity-functional/pom.xml
```diff
@@ -31,10 +31,12 @@
   <description>BDD tests for AET</description>
 
   <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
```
subiron: do we need this extra line?

wiiitek: Nice to have:
* https://maven.apache.org/general.html#encoding-warning

PR #55, opened 2017-02-05
Introducing AET changelog.

Skejven:
Here is a pull request with an idea to track significant changes through the AET history.
The idea is to update `CHANGELOG.md` file every time an important pull request is merged into master so that we can easily show what bugfixes/enchancments/new features are delivered within new version.
Contribution and feedback are welcome.
___

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+![Cognifide logo](http://cognifide.github.io/images/cognifide-logo.png)
+<p align="center">
+  <img src="https://github.com/Cognifide/aet/blob/master/misc/img/aet-logo-black.png?raw=true"
+         alt="AET Logo"/>
+</p>
+
 # AET Releases
 All notable changes to AET will be documented in this file.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,18 @@
+# AET Releases
+All notable changes to AET will be documented in this file.
+
+## Unreleased
+**List of changes that are finished but not yet released in any final version.**
+- [PR-41](https://github.com/Cognifide/aet/pull/41) - [Click](https://github.com/Cognifide/aet/wiki/ClickModifier)
+ and [Hide](https://github.com/Cognifide/aet/wiki/HideModifier) modifiers allows to use css selectors as parameters. 
+ Added [Replace Text](https://github.com/Cognifide/aet/wiki/ReplaceTextModifier) modifier and it's tests.
+- [PR-39](https://github.com/Cognifide/aet/pull/39) - fixed filterint w3cHtml issues.
+
+
+## Version 2.0.1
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -41,6 +41,11 @@ This documentation after update is ported to [AET wiki](https://github.com/Cogni
 When updating documentation please update proper markdown pages in [documentation](https://github.com/Cognifide/aet/tree/master/documentation) module following [instructions](https://github.com/Cognifide/aet/blob/master/documentation/README.md) and include it with your pull request.
 After your pull request is merged, wiki pages will be updated.
 
+## Changelog
+When changing or fixing some important part of AET, please remember to update [Changelog](https://github.com/Cognifide/aet/blob/master/CHANGELOG.md).
+Your entry should be enlisted in `Unreleased` section. It will be moved to appropriate release notes when released.
+Please use convention `- [PR-ABC](https://github.com/Cognifide/aet/pull/ABC) - short description of the PR.`.
```
wiiitek: Maybe `[PR-123]` would be better for the convention example? What do you think?

pom.xml
```diff
@@ -623,14 +623,13 @@
...
               <!-- Files that don't require a license header -->
-              <exclude>**/README.md</exclude>
               <exclude>**/LICENSE</exclude>
               <exclude>**/NOTICE</exclude>
-              <exclude>CONTRIBUTING.md</exclude>
               <exclude>.travis.yml</exclude>
 
               <exclude>**/target/**</exclude>
 
+              <exclude>**/*.md</exclude>
```
wiiitek: If we are to exclude all `.md` files from rat plugin then we could also remove the exclusions from **documentation** module pom.

CHANGELOG.md
```diff
@@ -0,0 +1,24 @@
...
+
+# AET Releases
+All notable changes to AET will be documented in this file.
+
+## Unreleased
+**List of changes that are finished but not yet released in any final version.**
+- [PR-41](https://github.com/Cognifide/aet/pull/41) - [Click](https://github.com/Cognifide/aet/wiki/ClickModifier)
+ and [Hide](https://github.com/Cognifide/aet/wiki/HideModifier) modifiers allows to use css selectors as parameters. 
+ Added [Replace Text](https://github.com/Cognifide/aet/wiki/ReplaceTextModifier) modifier and it's tests.
+- [PR-39](https://github.com/Cognifide/aet/pull/39) - fixed filterint w3cHtml issues.
```
subiron: typo "filterint"

CONTRIBUTING.md
Skejven: I'd rather not put numbers here. Is `ABC` not clear?

pom.xml
Skejven: Good point, removed.

PR #82, opened 2017-03-20
doc updates

RdkLwg:
doc updates inc.: Collectors.md, Comparators.md, CookieComparator.md, CookieModifier.md, DataFilters.md, DatabaseStructure.md, DefiningSuite.md, Dictionary.md

### DIFF
--- a/documentation/src/main/wiki/Dictionary.md
+++ b/documentation/src/main/wiki/Dictionary.md
@@ -4,10 +4,10 @@
 a *JMS (Java Message Service)* Server which is a basic communication channel between AET System components.
 
 *AET*  
-an acronym for **A**utomatic **E**xploratory **T**esting, a testing tool developed by Cognifide.
+an open source testing tool developed by Cognifide.
 
 *AET Core*  
-a set of system modules that are crucial to whole system work. The AET system will not work properly without all core modules configured and running properly.
+a set of system modules that are crucial to the whole system to work. The AET system will not work properly without all core modules configured and running properly.
 
 *AET Jobs*  
 implementations of jobs that can perform a particular task (e.g. collect screenshots, compare sources, validate a page against *W3C HTML5*).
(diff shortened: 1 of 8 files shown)


documentation/src/main/wiki/Dictionary.md
```diff
@@ -4,10 +4,10 @@
 a *JMS (Java Message Service)* Server which is a basic communication channel between AET System components.
 
 *AET*  
-an acronym for **A**utomatic **E**xploratory **T**esting, a testing tool developed by Cognifide.
+an open source testing tool developed by Cognifide.
 
 *AET Core*  
-a set of system modules that are crucial to whole system work. The AET system will not work properly without all core modules configured and running properly.
+a set of system modules that are crucial to the whole system to work. The AET system will not work properly without all core modules configured and running properly.
```
toniedzwiedz: > crucial to the whole system to work

sounds a bit strange. 

How about

> crucial for the whole system to work

or

> crucial to the operation of the whole system

I'm just following a gut feeling here, struggling to find some clear guidelines by a credible source.

documentation/src/main/wiki/Dictionary.md
```diff
@@ -28,7 +28,7 @@ usually used in the context of a small piece of data, the result of some operati
 see *Amazon Web Services*.
 
 *Baseline*  
-The act of taking a snapshot of the url/page and saving it to a file for future comparison in a number of ways to find differences.
+the act of taking a snapshot of the url/page and saving it to a file for its future comparison in a number of ways to find differences.
```
toniedzwiedz: Still doesn't sound quite right. Correct me if I'm wrong but _baseline_, as a noun, is not the act of taking a snapshot. It's the snapshot itself. 

If, on the other hand, we're liberal enough to use the word _baseline_  as a verb, I believe _the act of taking a snapshot_ would be... _baselining_? I suggest we just stick with the noun :)

How about: _a snapshot of the page, stored in a file, against which future versions will be compared._

Skejven:
Fixes introduced.

PR #85, opened 2017-04-11
execution time stats

Skejven:
Collecting some execution time stats and present them on the report. Later it may be used to analyse time consumed for processing the suite.
___

### DIFF
--- a/api/communication-api/src/main/java/com/cognifide/aet/communication/api/metadata/Statistics.java
+++ b/api/communication-api/src/main/java/com/cognifide/aet/communication/api/metadata/Statistics.java
@@ -1,26 +1,34 @@
+/**
+ * Automated Exploratory Tests
+ *
+ * Copyright (C) 2013 Cognifide Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


api/communication-api/src/main/java/com/cognifide/aet/communication/api/metadata/Statistics.java
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.cognifide.aet.communication.api.metadata;
+
+import java.io.Serializable;
+
+public class Statistics implements Serializable {
```
subiron: Is there anything else that we can measure inside Statistics class in the future?
I'm asking because it looks like we have here the whole class to wrap single field with no additional reasons.

Skejven: Yes, my idea was to add more data here in the future, e.g. start/end time if required, maybe some wait/sleep time required to pass.
I'd like to have statistics of how long AET really needed to perform a test (how long loading/sleep took vs real test time).

Skejven:
@subiron measuring full statistics at `Test` and `Url` level is very trick and I didn't include it in this pull request.
The problem is that test and url processing (I mean both collect and compare phases) is performed in dispersed environment. AET uses (especially during comparison phase) idea similar to `divide and conquer`. This mean totally separate workers/nodes execute parts of test and url processing. Because of that storing `Statistics` for the url or test may even not make a sens - it would be better to sum up all statistics of descendant data nodes. But this can be already done with the data that is already stored in those nodes (for url, these nodes are `steps`, for test it would be the sum of all `steps` of all `urls`).

PR #89, opened 2017-05-04
improve & refactor filters 

subiron:
This is just first part  (code)  of whole changes, it's meant to accelerate CR.
TBD
- sanity suite update
- functional tests update (if needed)

### DIFF
--- a/core/jobs/src/main/java/com/cognifide/aet/job/common/datafilters/jserrorsfilter/JsErrorsFilter.java
+++ b/core/jobs/src/main/java/com/cognifide/aet/job/common/datafilters/jserrorsfilter/JsErrorsFilter.java
@@ -17,32 +17,30 @@
  */
 package com.cognifide.aet.job.common.datafilters.jserrorsfilter;
 
-import com.google.common.base.Predicate;
-import com.google.common.collect.FluentIterable;
-
 import com.cognifide.aet.job.api.collector.JsErrorLog;
 import com.cognifide.aet.job.api.datafilter.AbstractDataModifierJob;
 import com.cognifide.aet.job.api.exceptions.ParametersException;
 import com.cognifide.aet.job.api.exceptions.ProcessingException;
-
-import org.apache.commons.lang3.StringUtils;
-
+import com.google.common.base.Predicate;
+import com.google.common.collect.FluentIterable;
 import java.util.Map;
 import java.util.Set;
-
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 import javax.annotation.Nullable;
+import org.apache.commons.lang3.StringUtils;
 
 public class JsErrorsFilter extends AbstractDataModifierJob<Set<JsErrorLog>> {
 
   public static final String NAME = "js-errors-filter";
 
-  private static final String PARAM_ERROR = "error";
+  private static final String PARAM_ERROR = "errorPattern";
 
   private static final String PARAM_SOURCE = "source";
 
   private static final String PARAM_LINE = "line";
 
-  private String errorMessage;
+  private Pattern errorMessagePattern;
 
   private String sourceFile;
 
@@ -50,17 +48,23 @@ public class JsErrorsFilter extends AbstractDataModifierJob<Set<JsErrorLog>> {
 
   @Override
   public void setParameters(Map<String, String> params) throws ParametersException {
-    errorMessage = params.get(PARAM_ERROR);
+    if (params.containsKey(PARAM_ERROR)) {
+      try {
+        errorMessagePattern = Pattern.compile(params.get(PARAM_ERROR));
+      } catch (PatternSyntaxException e) {
+        throw new ParametersException("errorPattern value is invalid regular-expression pattern.", e);
+      }
+    }
     sourceFile = params.get(PARAM_SOURCE);
     if (params.containsKey(PARAM_LINE)) {
       try {
         line = Integer.parseInt(params.get(PARAM_LINE));
       } catch (NumberFormatException e) {
         throw new ParametersException(
-                "Provided line: " + params.get(PARAM_LINE) + " is not a numeric value.", e);
+            "Provided line: " + params.get(PARAM_LINE) + " is not a numeric value.", e);
       }
     }
-    validateParameters(errorMessage, sourceFile, line);
+    validateParameters(errorMessagePattern, sourceFile, line);
   }
 
   @Override
@@ -75,25 +79,25 @@ public class JsErrorsFilter extends AbstractDataModifierJob<Set<JsErrorLog>> {
 
   private boolean shouldFilterOut(JsErrorLog jse) {
     return matchSourceFile(sourceFile, jse.getSourceName())
-            && matchStrings(errorMessage, jse.getErrorMessage())
-            && matchNumbers(line, jse.getLineNumber());
+        && matchPattern(errorMessagePattern, jse.getErrorMessage())
+        && matchNumbers(line, jse.getLineNumber());
   }
 
   @Override
   public String getInfo() {
-    return NAME + " DataModifier with parameters: " + PARAM_SOURCE + ": " + sourceFile + " " + PARAM_ERROR
-            + ": " + errorMessage + " " + PARAM_LINE + ": " + line;
+    return NAME + " DataModifier with parameters: " + PARAM_SOURCE + ": " + sourceFile + " "
+        + PARAM_ERROR + ": " + errorMessagePattern + " " + PARAM_LINE + ": " + line;
   }
 
-  private void validateParameters(String errorMessege, String sourceFile, Integer line)
-          throws ParametersException {
+  private void validateParameters(Pattern errorMessege, String sourceFile, Integer line)
+      throws ParametersException {
     if (errorMessege == null && sourceFile == null && line == null) {
       throw new ParametersException("At least one parameter must be provided");
     }
   }
 
-  private boolean matchStrings(String paramValue, String errorValue) {
-    return StringUtils.isEmpty(paramValue) || StringUtils.equalsIgnoreCase(paramValue, errorValue);
+  private boolean matchPattern(Pattern errorMessagePattern, String errorValue) {
+    return errorMessagePattern == null ||  errorMessagePattern.matcher(errorValue).matches();
   }
 
   private boolean matchSourceFile(String paramValue, String errorValue) {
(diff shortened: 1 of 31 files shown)


documentation/src/main/wiki/releases/2.0.0/Documentation-2.0.0.md
```diff
@@ -2556,7 +2556,7 @@ There can be more than one `accessibility-filter` tag in `accessibility` compara
 
 #### Extranct Element Data Filter
```
joannajasnowska: typo

core/jobs/src/main/java/com/cognifide/aet/job/common/datafilters/jserrorsfilter/JsErrorsFilter.java
```diff
@@ -17,50 +17,40 @@
...
-
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
 public class JsErrorsFilter extends AbstractDataModifierJob<Set<JsErrorLog>> {
 
   public static final String NAME = "js-errors-filter";
 
-  private static final String PARAM_ERROR = "error";
+  private static final String PARAM_ERROR = "errorPattern";
```
Skejven: I'm not sure this is good idea to change params names such drastically.
IMO we should give users at least one release cycle to adjust otherwise the old suites won't work with released version.
We definitely should mention such changes in some kind `migration log`.

Just a thought - maybe we should provide some kind of service that will automatically adjust old suites to new form?

core/jobs/src/main/java/com/cognifide/aet/job/common/datafilters/removeregexp/ReplaceRegexpDataModifier.java
```diff
@@ -68,13 +74,14 @@ public String modifyPattern(String data) throws ProcessingException {
...
+    return NAME + " DataModifier with parameters: " + DATA_REGEXP + ": " + dataRegExp + " " + PATTERN_REGEXP + ": "
+        + patternRanges;
   }
 
-  private String modify(String data, String regexp) throws ProcessingException {
+  private String modify(String data, Pattern pattern) throws ProcessingException {
     String result = null;
     try {
-      result = data.replaceAll(regexp, StringUtils.EMPTY);
+      result = pattern.matcher(data).replaceAll(replacement);
```
Skejven: If user won't pass `value` argument, then `replacement` is `null`. Is that ok in this case?

core/jobs/src/main/java/com/cognifide/aet/job/common/utils/ParamsHelper.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathExpression;
+import javax.xml.xpath.XPathExpressionException;
+import javax.xml.xpath.XPathFactory;
+import org.apache.commons.lang3.ObjectUtils;
+
+public class ParamsHelper {
```
Skejven: I like idea of this helper very much. It looks that lot of code is now simpler and redundancy is limited.
Please consider enriching this class's methods with javadoc, I think that some of this code may be reused in many future modifiers.

I'm also wondering if following scenarios:

1. Don't return `null`, use `Optional` instead.
2. Don't return `null`, provide option with 3rd - `default` parameter and return it when no param with key provided.

core/jobs/src/test/java/com/cognifide/aet/job/common/datafilters/statuscodesfilter/ExcludeStatusCodesFilterTest.java
```diff
@@ -34,20 +34,11 @@
 @RunWith(MockitoJUnitRunner.class)
 public class ExcludeStatusCodesFilterTest extends StatusCodesFilterTestBase {
 
-  @Test
```
Skejven: Why so many unit tests were removed? Shouldn't we replace them if logic changed?

core/jobs/src/main/java/com/cognifide/aet/job/common/utils/ParamsHelper.java
```diff
@@ -0,0 +1,112 @@
...
+      } catch (NumberFormatException e) {
+        throw new ParametersException(
+            "Provided line: " + params.get(result) + " is not a numeric value.", e);
+      }
+    }
+    return result;
+  }
+
+  public static String getParamAsString(String key, Map<String, String> params) {
+    String result = null;
```
joannajasnowska: method can be shorter if you use ternary operator

core/jobs/src/main/java/com/cognifide/aet/job/common/datafilters/removeregexp/ReplaceRegexpDataModifier.java
subiron: fixed

subiron:
@Skejven please take a look again. Current changes in filters only extends existing behavior instead of modifying it (there is no need to change existing suite definitions).

documentation/src/main/wiki/JSErrorsDataFilter.md
```diff
@@ -15,7 +15,13 @@ Resource name: js-errors
 | --------- | ----- | ----------- | --------- |
 |`error`|string error|Exact error message|At least one of parameter is required|
 |`source`|string file name|Source file name (full path including `http://`) in which error occurred|
-|`line`|integer line number|Line number in file in which error occurred|
+|`errorPattern` | pattern error text | Regular expression that matches message text of issue to be filter out|At least one parameter is required|
+|`sourcePattern` | pattern file name | Regular expression that should match file name to be filter out (full path including `http://`) in which error occurred|
+|`line`  integer line number|Line number in file in which error occurred|
+*Note:
```
Skejven: Looks like not closed `*`.

documentation/src/main/wiki/RemoveRegexDataFilter.md
```diff
@@ -1,20 +1,21 @@
-#### Remove Regex Data Filters
```
Skejven: We probably should also change the name of this file.

pom.xml
```diff
@@ -163,7 +163,7 @@
       <dependency>
         <groupId>org.apache.commons</groupId>
         <artifactId>commons-lang3</artifactId>
-        <version>3.3.2</version>
+        <version>3.5</version>
```
Skejven: Please upgrade the version also in `aet-features.xml` file.

core/jobs/src/main/java/com/cognifide/aet/job/common/utils/ParamsHelper.java
```diff
@@ -0,0 +1,160 @@
...
+
+public class ParamsHelper {
+
+  /***
+   * @param key property name that should store regexp
+   * @param params map of parameters
+   * @return Pattern  or null if there is no key in params map
+   * @throws ParametersException if Pattern under key is provided but invalid
+   */
+  public static Pattern getAsPattern(String key, Map<String, String> params) throws ParametersException {
```
Skejven: How about one of the the ideas for all this helper methods:

1. Don't return null, use Optional instead.
2. Don't return null, provide option with 3rd - default parameter and return it when no param with key provided.

core/jobs/src/main/java/com/cognifide/aet/job/common/utils/ParamsHelper.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathExpression;
+import javax.xml.xpath.XPathExpressionException;
+import javax.xml.xpath.XPathFactory;
+import org.apache.commons.lang3.ObjectUtils;
+import org.apache.commons.lang3.StringUtils;
+
+public class ParamsHelper {
```
Skejven: Since this class looks like Helper/Util class please mark it as one (create private constructor and mark this class as `final`).

core/jobs/src/main/java/com/cognifide/aet/job/common/datafilters/statuscodesfilter/StatusCodesFilter.java
```diff
@@ -22,52 +22,37 @@
 import com.cognifide.aet.job.api.exceptions.ProcessingException;
 import com.cognifide.aet.job.common.collectors.statuscodes.StatusCode;
 import com.cognifide.aet.job.common.collectors.statuscodes.StatusCodesCollectorResult;
```
wiiitek: Wrong import order! Should be as in other JAVA files (com, org, java).

core/jobs/src/main/java/com/cognifide/aet/job/common/utils/ParamsHelper.java
```diff
@@ -0,0 +1,160 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.cognifide.aet.job.common.utils;
+
+
+import com.cognifide.aet.job.api.exceptions.ParametersException;
```
wiiitek: Although those imports are OK according to Google style guide, we have it different in other places.
Which version do we want to have?

Skejven: Google style guide changed imports settings for IntelliJ in [one of the last updates](https://github.com/google/styleguide/commit/75c289f1d33836d1ff4bd94e6c9033673e320b58). It may occur that we have some formatted with the old and some with the new version.

wiiitek: Wow! Thank you @Skejven for noticing that!

core/jobs/src/main/java/com/cognifide/aet/job/common/utils/ParamsHelper.java
wiiitek: Done

pom.xml
wiiitek: Not needed anymore.

documentation/src/main/wiki/RemoveRegexDataFilter.md
wiiitek: Not needed anymore.

core/jobs/src/main/java/com/cognifide/aet/job/common/utils/ParamsHelper.java
wiiitek: This method was moved inside `getPatternFromPatternParameterOrPlainText`.

documentation/src/main/wiki/JSErrorsDataFilter.md
wiiitek: Fixed

wiiitek:
New pull request created with smaller set of changes.
See:
* https://github.com/Cognifide/aet/pull/97

PR #90, opened 2017-05-05
Zgcq 7940 aet deploy relesed version

KAJUNIEK:
Add gradle configuration for release deploy
kill karaf, clear cache, deploy artifacts, start karaf again on all <env(integration, staging production)> instances
(dig up gradle scripts from legacy aet repo, addtional setup of ssh server may be needed on prod or stg)
upload report war into report container app for given <env>


subiron:
please remove generated files and add them to git ignore

PR #97, opened 2017-05-09
Support for regular expressions...

wiiitek:


core/jobs/src/main/java/com/cognifide/aet/job/common/utils/ParamsHelper.java
```diff
@@ -0,0 +1,122 @@
...
+
+  public static String getParamAsString(String key, Map<String, String> params) {
+    return params.containsKey(key) ? params.get(key) : null;
+  }
+
+  /***
+   * @param values array of objects
+   * @throws ParametersException if all passed objects are null
+   */
+  public static void atLeastOneIsProvided(Object... values)
```
Skejven: How about: [`ObjectUils.firstNotNull`](http://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/ObjectUtils.html#firstNonNull%28T...%29) `!= null`?

If you don't like it, please just least `return` instead line `70` and throw the exception without checking `allNull`.

documentation/src/main/wiki/releases/2.0.0/Documentation-2.0.0.md
```diff
@@ -2554,9 +2554,9 @@ There can be more than one `accessibility-filter` tag in `accessibility` compara
 ```
 
 
-#### Extranct Element Data Filter
+#### Extract Element Data Filter
```
Skejven: Nice one :)

documentation/src/main/wiki/SourceCollector.md
```diff
@@ -6,7 +6,7 @@ Module name: **source**
 
 | ! Note    |
 | :-------- |
-| System is waiting up to 20 seconds before request is timed out. This parameter is configurable via OSGi configuration (`AET Source Collector Factory`). |
+| System is waiting up to 40 seconds before request is timed out. This parameter is configurable via OSGi configuration (`AET Source Collector Factory`). |
```
Skejven: Are you sure that it is switched to 40 seconds?
It looks like you forgot to update [SourceCollectorFactory](https://github.com/Cognifide/aet/blob/master/core/jobs/src/main/java/com/cognifide/aet/job/common/collectors/source/SourceCollectorFactory.java) `DEFAULT_TIMEOUT` constant.

wiiitek: Thank you! I don't ant to update defaults in this pull requests. If longer timeout is needed it should be done by OSGi configuration (conf files). I will revert this docs to 20 seconds..

core/jobs/src/main/java/com/cognifide/aet/job/common/utils/ParamsHelper.java
wiiitek: Thanks, I will rewrite this.

core/jobs/src/main/java/com/cognifide/aet/job/common/datafilters/jserrorsfilter/JsErrorsFilter.java
```diff
@@ -74,34 +72,21 @@ public boolean apply(@Nullable JsErrorLog input) {
...
 
   private boolean shouldFilterOut(JsErrorLog jse) {
-    return matchSourceFile(sourceFile, jse.getSourceName())
-            && matchStrings(errorMessage, jse.getErrorMessage())
-            && matchNumbers(line, jse.getLineNumber());
+    boolean sourceNotSpecified = sourceFile == null;
+    boolean sourceEquals = jse.getSourceName().equals(sourceFile);
+    boolean sourceEndsWith = jse.getSourceName().endsWith("/" + sourceFile);
+
+    return (sourceNotSpecified || sourceEquals || sourceEndsWith)
```
subiron: please allow adding the ability to ignore specific source file not all files with the same name.
eg:
index.html have linked following scripts:
http://my-page.com/data/jquery.js
http://thirdparty.com/amazon/jquery.js
http://thirdparty.com/amazon/version/2/jquery.js
http://thirdparty.com/amazon/version/2-min/jquery.js
and a single error occur in "http://thirdparty.com/amazon/version/2-min/jquery.js"

currently, a user is unable to ignore just "http://thirdparty.com/amazon/version/2-min/jquery.js"

wiiitek: It is possible by specifying a long path. For example this filter:
```xml
<js-errors-filter source="comparators/jserrors/jsFileWithError.js" />
```
is working and will not hide errors from `some/other/path/jsFileWithError.js`
Is it OK or do we need some changes here?

subiron: ah ok,  I had to skip "sourceEndsWith" part, sorry

PR #103, opened 2017-05-17
Updates for source param for JS errors

wiiitek:
Update for source param to work with folloowing configs:
```xml
<js-errors-filter source="/sample-site/sanity/comparators/jserrors/failed.js"/>
<js-errors-filter source="/js/table.js"/>
<js-errors-filter source=".js"/>
```


core/jobs/src/main/java/com/cognifide/aet/job/common/datafilters/jserrorsfilter/JsErrorsFilter.java
```diff
@@ -72,15 +72,24 @@ public boolean apply(@Nullable JsErrorLog input) {
...
   }
 
+  private boolean shouldExcludeRegardingSource(String errorSource) {
+    boolean shouldExclude;
+
+    boolean sourceParamNotSpecified = sourceFile == null;
+    if (sourceParamNotSpecified) {
+      shouldExclude = true;
+    } else {
+      shouldExclude = errorSource.equals(sourceFile) || errorSource.endsWith(sourceFile);
```
subiron: `errorSource.endsWith(sourceFile);`
will return true if sourceFile is an empty string,
imo this is not expected behavior,
 maybe check if sourceFile isBlank in line 84 instead of null check.

PR #107, opened 2017-05-18
Update for another suite patterns branch

wiiitek:
Some code changes after feature testing


core/datastorage/src/main/java/com/cognifide/aet/vs/metadata/DocumentConverter.java
```diff
@@ -29,6 +29,7 @@
 class DocumentConverter {
 
   private static final Gson GSON_FOR_MONGO_JSON = new GsonBuilder()
+      //.registerTypeHierarchyAdapter(Optional.class, new OptionalDeserializer())
```
subiron: why do we want to leave this line here?

wiiitek: To be removed, thanks!

pom.xml
```diff
@@ -601,6 +601,9 @@
           <groupId>com.cognifide.aet</groupId>
           <artifactId>aet-maven-plugin</artifactId>
           <version>${project.version}</version>
+          <configuration>
```
subiron: is it needed?

wiiitek: yes, it is used when we want to specify different pattern suite from command line.
then we overwrite this empty value with:
```
-DpatternSuite=demo
```

PR #111, opened 2017-05-22
Shared patterns feature - pattern id validation

wiiitek:


test-executor/src/main/java/com/cognifide/aet/executor/SuiteValidator.java
```diff
@@ -94,4 +119,21 @@ private static boolean hasScreenComparator(TestRun testRun) {
...
+    SimpleDBKey dbKey = new SimpleDBKey(testSuiteRun.getCompany(), testSuiteRun.getProject());
+    String correlationId = testSuiteRun.getCorrelationId();
+    Suite patternSuite = null;
+    try {
+      patternSuite = metadataDAO.getSuite(dbKey, correlationId);
+    } catch (StorageException se) {
+      LOG.error("error while retrieving suite from mongo db: '{}', correlationId: '{}'",
+          dbKey, correlationId, se);
+    }
+    if (patternSuite!=null) {
```
Skejven: Please fix spacing here to:
`(patternSuite != null)`.

Or even better, how about:
`return patternSuite != null` ?

PR #112, opened 2017-05-22
Support for using patterns of another suite

wiiitek:


api/communication-api/src/main/java/com/cognifide/aet/communication/api/metadata/Suite.java
```diff
@@ -74,6 +92,8 @@
   @Valid
   private final List<Test> tests = new ArrayList<>();
 
+  private String patternCorrelationId;
```
szymon-owczarzak-cognifide: Can this be final? Is it ok that it can be changed?

wiiitek: Thanks! I will change it to final :)

Skejven: I don't think it can be final. It is an optional parameter which is not set in constructor.

szymon-owczarzak-cognifide: So maybe we need new constructor or factory method.

wiiitek: Made final as agreed with @Skejven

PR #150, opened 2017-10-01
fix for incorrect number of arguments for loggers

wiiitek:


core/worker/src/main/java/com/cognifide/aet/worker/listeners/ComparatorMessageListenerImpl.java
```diff
@@ -87,8 +87,11 @@ public void onMessage(final Message message) {
     String jmsCorrelationId = JmsUtils.getJMSCorrelationID(message);
 
     if (comparatorJobData != null && StringUtils.isNotBlank(jmsCorrelationId)) {
-      LOGGER.info("ComparatorJobData  [{}] message arrived. CorrelationId: {} TestName: {} UrlName: {}",
```
wiiitek: Should have 4 arguments.

core/worker/src/main/java/com/cognifide/aet/worker/listeners/ComparatorMessageListenerImpl.java
```diff
@@ -99,8 +102,11 @@ public void onMessage(final Message message) {
                 .newBuilder(comparatorJobData.getTestName(), comparatorJobData.getUrlName(), step.getIndex());
         try {
           Comparator processedComparator = dispatcher.run(comparator, properties);
-          LOGGER.info("Comparison successfully ended. CorrelationId: {} TestName: {} Comparator: {}",
```
wiiitek: 4 arguments provided, so we need to add placeholder in log message.

PR #158, opened 2017-10-28
Changes for easier upgrade to new karaf

wiiitek:
___

### DIFF
--- a/osgi-dependencies/w3chtml5validator/pom.xml
+++ b/osgi-dependencies/w3chtml5validator/pom.xml
@@ -33,6 +33,11 @@
   <name>AET :: OSGi Dependencies :: HTML validator</name>
   <description>Wraps nu.validator library into OSGi bundle</description>
 
+  <properties>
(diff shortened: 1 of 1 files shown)


osgi-dependencies/aet-features.xml
```diff
@@ -47,6 +47,7 @@
         <bundle>mvn:commons-collections/commons-collections/3.2.1</bundle>
         <bundle>mvn:commons-fileupload/commons-fileupload/1.3.2</bundle>
         <bundle>mvn:commons-io/commons-io/2.4</bundle>
+        <bundle>mvn:commons-lang/commons-lang/2.6</bundle>
```
Skejven: Is that new lib? Where it is or will be used?

wiiitek: It is not new: it is currently available on Karaf, but on the new one it is missing.
Putting it here doesn't change anything for current version.

Skejven: Yes, I think this is a good idea if the dependency is only in AET code.

PR #69, opened 2017-03-02
JsModifier - Running JavaScript command as in the browser console.

pnad:
JsModifier - Running JavaScript command as in the browser console.

PR #164, opened 2017-11-19
Execute js modifier supports external snippets

Skejven:
According to the discussion at #69 - `<executejavascript>` can now fetch js snippets from the external source (and authenticate with basic auth ther if needed).
___


wiiitek:
___

PR #169, opened 2017-11-26
Deprecated resolution modifier's parameter maximize removed, height no longer mandatory

Skejven:
I removed deprecated `maximize` param from resolution modifier.
Additionally, `height` parameter is no longer mandatory.
___


Skejven:
This will be done in #294

PR #170, opened 2017-11-26
documentation updates

RdkLwg:
Documentation updates

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
          alt="AET Logo"/>
 </p>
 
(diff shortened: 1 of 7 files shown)


wiiitek:
Thank you @RdkLwg !

README.md
```diff
@@ -10,7 +10,7 @@
          alt="AET Logo"/>
 </p>
 
-AET (acronym formed from **A**utomated **E**xploratory **T**esting) is a system that detects changes on web sites.
+AET is a system that detects changes on web sites.
```
Skejven: I'm just wondering. Maybe we should add here the info that AET also performs basic page health check (like w3c compliance, accessibility etc.)?

PR #175, opened 2017-12-15
DO NOT MERGE!!! delta changes for wait-for modifiers (partial code review)

wiiitek:


integration-tests/sample-site/src/main/webapp/WEB-INF/web.xml
```diff
@@ -53,6 +53,16 @@
         <role-name>SAMPLE_ROLE</role-name>
     </security-role>
 
+    <filter>
+        <filter-name>3 Seconds Latency</filter-name>
```
Skejven: Isn't that 4 seconds latency?

wiiitek: You're right! thanks!

Issue #165, opened 2017-11-21
Ambiguous message presented when test name contain /


PiotrRybickiCognifide:
The message presented by Maven Client that runned suite with Test containing / in the name doesn't say the reason correctly:

> [ERROR] ... Failed to process test suite: Not Found -> [Help 1]

Message should be:
>[ERROR] At least one of the Tests in the suite contain unsupported character. Please check and rename tests containing special characters.<

mchrominski:
Currently, after the changes in #183 the message looks like:
`[ERROR] Failed to schedule test suite: '[Status: 400] Something is wrong with your suite definition! Value 'first/test' of attribute name in test does not match pattern `[a-zA-Z\-_0-9]+`!'`

I believe this is self-exploratory now so closing this issue.

Issue #161, opened 2017-11-13
Ambiguous message presented when suite is locked


Skejven:
The message presented by Maven Client that runned suite which is currently locked (eg. actually running) doesn't say the reason correctly:
```
[ERROR] Failed to process test suite: 'Not Found'
```

The message should be rather like:

```
[ERROR] This suite is currently running on this AET instance. Please wait until it finishes before scheduling new run.
```

PR #183, opened 2017-12-29
Fix ambiguous api communicates

Skejven:
Fixes ambiguous messages from AET API (#161 and #165).
___


test-executor/src/main/java/com/cognifide/aet/executor/SuiteExecutor.java
```diff
@@ -17,46 +17,44 @@
  */
 package com.cognifide.aet.executor;
```
wiiitek: Maybe we could leave the imports order for now and do the change for all classes in separate pull request?

Skejven: @wiiitek, thanks for comments and update. Let's keep this order of imports here. I will update imports order in the codebase and propose a consistent formatting in the repo in a few days (and update the docs).

Skejven:
@wiiitek I've uploaded currently used google formatting configurations in #192 and created PR with reformatted imports in #193 .

PR #195, opened 2018-01-05
Reformat java code according to style

Skejven:
I've reviewed all reformatted classes. I didn't reformat code of classes that were copied form other AL2.0 projects (only our license header).
___


Skejven:
We will go with #196 instead.

PR #201, opened 2018-01-14
fixing issue in login modifier documentation and fixing formating of …

Danon9111:
…error message

### DIFF
--- a/documentation/src/main/wiki/LoginModifier.md
+++ b/documentation/src/main/wiki/LoginModifier.md
@@ -16,7 +16,7 @@ Module name: **login**
 | --------- | ----- | --------- | ------------- |
 | `login` | User's login | no | admin |
 | `password` | Password | no | admin |
-| `login-page` | Url to login page | yes | [http://localhost:4502/libs/granite/core/content/login.html](http://localhost:4502/libs/granite/core/content/login.html) |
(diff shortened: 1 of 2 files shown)


documentation/src/main/wiki/LoginModifier.md
```diff
@@ -16,7 +16,7 @@ Module name: **login**
 | --------- | ----- | --------- | ------------- |
 | `login` | User's login | no | admin |
 | `password` | Password | no | admin |
-| `login-page` | Url to login page | no | [http://localhost:4502/libs/granite/core/content/login.html](http://localhost:4502/libs/granite/core/content/login.html) |
+| `login-page` | Url to login page | yes | [http://localhost:4502/libs/granite/core/content/login.html](http://localhost:4502/libs/granite/core/content/login.html) |
```
Skejven: Thank you @Danon9111 for spotting this.
I think also the default value is not right here - as I searched [`LoginModifierConfig`](https://github.com/Cognifide/aet/blob/master/core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/login/LoginModifierConfig.java) there is no such value.
Could you please also remove it from the doc?

Thanks

wiiitek: I have removed that :)

PR #202, opened 2018-01-15
versions update

joannajasnowska:
VirtualBox version updated


Skejven:
Could you please also update the `CHANGELOG.md` and `WhatsNew`?
This is the change that users should be aware of.
Thanks

PR #205, opened 2018-01-17
Zgcq 9547 aet features xml generated by maven

dzuma:


osgi-dependencies/pom.xml
```diff
@@ -40,6 +40,36 @@
     <module>w3chtml5validator</module>
   </modules>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-resources-plugin</artifactId>
+        <executions>
+          <!-- process login modifier template to insert correct domain -->
```
szymon-owczarzak-cognifide: I think this comment could be deleted.

wiiitek: 👍  You're right. It is misleading here.

PR #216, opened 2018-01-30
ZGCQ-9485 remove border

edworczak:
## Description
There was an unnecessary border-bottom on last list element in each subgroup in sidepanel lists.
It has been removed.

---

### DIFF
--- a/report/src/main/webapp/assets/sass/_sidebar.scss
+++ b/report/src/main/webapp/assets/sass/_sidebar.scss
@@ -146,10 +146,6 @@
         margin-left: 15px;
         color: $gray_dark;
 
-        &:last-of-type {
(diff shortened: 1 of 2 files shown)


report/src/main/webapp/assets/sass/_sidebar.scss
```diff
@@ -147,7 +147,7 @@
         color: $gray_dark;
 
         &:last-of-type {
-          border-bottom: solid 1px #313942;
```
wiiitek: If this line is not needed let's remove it.
If someone wants it back we will search it in commit history.

Let's not have commented-out code.

edworczak: Lines are removed in a new commit.

PR #227, opened 2018-02-21
Refactoring

wiiitek:
Refactoring

## Description
Formatting change and argument copied so that argument is not modified.
The returned set [is later saved](https://github.com/Cognifide/aet/blob/master/core/jobs/src/main/java/com/cognifide/aet/job/common/comparators/jserrors/JsErrorsComparator.java#L73) by `artifactDao` so the AET system works as before.

## Motivation and Context
This is refactoring before changes that we want to introduce for better presentation of JS errors on AET report.

## Types of changes

## Checklist:

---


core/jobs/src/main/java/com/cognifide/aet/job/common/datafilters/jserrorsfilter/JsErrorsFilter.java
```diff
@@ -57,14 +58,27 @@ public void setParameters(Map<String, String> params) throws ParametersException
...
+   */
   @Override
-  public Set<JsErrorLog> modifyData(Set<JsErrorLog> data) throws ProcessingException {
-    return FluentIterable.from(data).filter(new Predicate<JsErrorLog>() {
-      @Override
-      public boolean apply(@Nullable JsErrorLog input) {
-        return !shouldFilterOut(input);
-      }
-    }).toSet();
+  public Set<JsErrorLog> modifyData(Set<JsErrorLog> data) {
```
Skejven: Since we have Java 8 maybe we could a `java.util` streams here instead?

wiiitek: That is true! I just wanted to introduce as little changes as possible..

szymon-owczarzak-cognifide: Exactly:
`return data.stream().filter(input -> !shouldFilterOut(input)).collect(Collectors.toSet());`

wiiitek: Yeah! There will be changes in near future that will do something similar, but without filtering ;)

PR #229, opened 2018-02-26
Note tooltip new

edworczak:
## Description
Removed empty space on popover when there is no comment.
Provided different logic and icons for comment and when there is no comment.
Introduced Font Awesome.
Improved buttons UX in editing comments.

---


report/src/main/webapp/assets/sass/_toolbar.scss
```diff
@@ -235,6 +235,6 @@
   display: inline-block;
 }
 
-.preformatted {
-  font-family: monospace;
```
michaldziubek: Is this a desired change?
I think that it can cause some regression resulting in few visual glitches. For example here:
2018-02-26_16h04_26
2018-02-26_16h04_44

report/src/main/webapp/assets/sass/_popups.scss
```diff
@@ -39,6 +39,18 @@
...
 
+  &-footer_flex {
+    & {
+      align-items: center;
+      display: flex;
+      justify-content: space-between;
+    }
+
+    button + button {
+      margin-right: 0px;
```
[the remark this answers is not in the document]
edworczak: I tried to add modal-footer_flex class to all popups, but there is one in report.html, which I couldn't get on the site. I left the styling (button + button) until someone can tell when this popup occurs.

report/src/main/webapp/assets/sass/_toolbar.scss
edworczak: This class is back in css now.

PR #238, opened 2018-04-17
Feature/winter edition

jakublamprecht:
Those changes add winter theme to AETs that is automatically triggered two weeks before Christmas and is turned off two weeks after.
## Description

Changed mechanism of adding classes of status icons. Added directive that triggers winter theme, changes logo and makes snow fall. Also updated some markup which could be simplified with the new way of adding icon classes.
## Motivation and Context

It gives AET reports appropriate look and feel during winter time which can possibly boost users mood :)
## Types of changes

## Checklist:

---


report/src/main/webapp/app/app.config.js
```diff
@@ -26,6 +26,7 @@ require.config({
     'angularAMD': '../assets/libs/angularAMD/angularAMD',
     'lodash': '../assets/libs/lodash/dist/lodash',
     'angular-bootstrap': '../assets/libs/angular-bootstrap/ui-bootstrap-tpls',
+    'snowfall': '../assets/libs/snowfall/snowfall.min',
```
jakublamprecht: Maybe there's a better way to include this that I didn't think about?

report/src/main/webapp/app/layout/main/suite/mainView.suite.view.html
```diff
@@ -26,17 +26,11 @@
...
-                <i class="fas fa-check"></i>
-              </div>
-              <div ng-if="test.getStatus() == 'warning'">
-                <i class="fas fa-exclamation-triangle"></i>
+              <div ng-if="test.getStatus() != 'rebased'" class="failed">
+                <i class="fas {{theme.statusClasses[test.getStatus()]}}"></i>
               </div>
               <div ng-if="test.getStatus() == 'rebased'">
-                <i class="fas fa-cloud-upload-alt"></i>
+                <i class="fas {{theme.statusClasses[test.getStatus()]}}"></i>
```
jakublamprecht: Those elements are required, so that when a case gets accepted/rejected the icon is redrawn - simply adding an element with {{theme.statusClasses[test.getStatus()]}} won't work, since the italic tags are replaced with svg tags by fontawesome.

report/src/main/webapp/app/components/winterEdition.directive.js
```diff
@@ -0,0 +1,61 @@
...
+      var DAY_MARGIN = 14;
+      var isChristmas = isInDateRange(TODAY, CHRISTMAS_DATE, DAY_MARGIN);
+
+      if (isChristmas) {
+        _.merge($scope.theme, WINTER_THEME);
+        snowFall.snow(document.body, {flakeCount: 200, maxSpeed: 20});
+      }
+    }
+
+    function isInDateRange(today, eventDate, dayMargin) {
```
wiiitek: I was thinking about renaming the function to contain word *near* instead of *range*.
Checking if date is in range would expect arguments such as: date and range.
We are doing it slightly different.  ->Would it be easier to read after such change?

report/src/main/webapp/app/components/winterEdition.directive.js
```diff
@@ -0,0 +1,61 @@
+/*
```
wiiitek: the folder `report/src/main/webapp/app/components` is the old one and not realy in sync with John Papa guide about folders for code. let's move this directive to some folder with the functional-like name.

we can create new folder below `.../app` called *themes*.

report/src/main/webapp/report.html
```diff
@@ -28,7 +28,7 @@
   <script defer src="assets/icons/fontawesome-all.js"></script>
 </head>
 
-<body aet-hide-popovers aet-keyboard-shortcuts aet-sidepanel ng-class="{ 'menu-expanded': sidebarExpanded }">
+<body aet-hide-popovers aet-keyboard-shortcuts aet-sidepanel aet-winter-edition ng-class="{ 'menu-expanded': sidebarExpanded }">
```
wiiitek: is this CSS class needed? why? Maybe it should be only added in December ?

jakublamprecht: discussed

PR #260, opened 2018-06-04
Upgrade to Karaf 4.2.0

wiiitek:
## Description
Karaf 4.2.0 works with both JAVA 8 and JAVA 9. For Active MQ v 5.15.x to work we need to provide Spring legacy repository for Aries blueprint. See also http://mail-archives.apache.org/mod_mbox/karaf-user/201805.mbox/%3C7bede6b9-d282-ddb5-52fc-dca74f698381@nanthrax.net%3E

After this is accepted we could update the cookbook to make new Karaf version default there.

## Motivation and Context
We want to use newest possible version of Karaf so that it is easier for us to upgrade other System components (i.e. JAVA).

## Types of changes

## Checklist:

---

### DIFF
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -24,7 +24,7 @@
 ENV["GEM_PATH"] = nil
 ENV["GEM_HOME"] = nil
 
-AET_VERSION = '2.1.5'
+AET_VERSION = '2.1.6-SNAPSHOT'
 
 VAGRANTFILE_API_VERSION = '2'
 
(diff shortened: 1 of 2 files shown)


vagrant/Vagrantfile
```diff
@@ -86,8 +86,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         }
       },
       "aet" => {
+        "base_link" => "https://www.kubiczak.pl/aet",
```
Skejven: How about using maven staging repository?
You could easily deploy snapshot version there without having custom links: https://oss.sonatype.org/content/groups/staging/

wiiitek: Maven artifacts are something different than "release artifacts" that we use in cookbook.

"Release artifacts":
- zip packages that groups similar AET stuff: bundles, configs, report app.

Maven artifacts:
- are not grouped in any archives.

We eould need adjust AET cookbook to use Maven artifacts only.
Or we could create new Maven artifacts with ZIP packaging for cookbook.
What do you think?

PR #265, opened 2018-06-21
Runner simplification refactoring

Skejven:
## Description
Curren AET Runner flow is very complicated with unnecessary creation of threads and Guice DI.

## Motivation and Context
Complicated AET Runner implementation makes it harder to understand it and improve. This is the first refactor in a series of changes that I'd like to apply to the AET Runner.
I've tried to introduce as few changes at is possible, there are areas to improve (starting with introducing Java 8 features).

## Types of changes

## Checklist:

- [ ] I have updated the documentation accordingly. (In progress...)

---


Skejven:
Documentation update in progress...

core/runner/src/main/java/com/cognifide/aet/runner/processing/TimeoutWatch.java
```diff
@@ -36,14 +32,14 @@ public synchronized void update() {
    * @param acceptedSecondsDifference accepted difference in seconds
    * @return true if last update was done more than acceptedSecondsDifference seconds from now
    */
-  public synchronized boolean isTimedOut(long acceptedSecondsDifference) {
```
wiiitek: Thank you for changes in visibility/scope :)

core/runner/src/main/java/com/cognifide/aet/runner/RunnerConfiguration.java
```diff
@@ -13,35 +13,26 @@
  * or implied. See the License for the specific language governing permissions and limitations under
```
wiiitek: Good changes. thanks!

core/runner/src/test/java/com/cognifide/aet/runner/processing/TimeoutWatchTest.java
```diff
@@ -1,3 +1,11 @@
+package com.cognifide.aet.runner.processing;
```
wiiitek: We usually have license header at the top of our files. could we have it consistent here?

core/runner/pom.xml
```diff
@@ -69,10 +69,6 @@
       <groupId>org.apache.activemq</groupId>
       <artifactId>activemq-osgi</artifactId>
     </dependency>
-    <dependency>
```
wiiitek: Awesome! Thank you @Skejven ! 

Actually it looks like guice is referenced also in *worker* module, but is not used at all there (or am I wrong?).
So we could remove Guice from the worker pom and then also from top-level pom, isn't it?

Then we could remove guice from `feature.xml` file. That would b great, right?

(there will be only some transitive dependencies to guice from our functional tests that use [Bobcat](https://github.com/Cognifide/bobcat)).

Could be done here or in another pull request if you prefer...

Skejven: Thank you @wiiitek for the review.
I will remove Guice dependency in the separate PR.

PR #268, opened 2018-06-29
Bobcat upgrade

wiiitek:
## Description
Upgrade to newest released version of Bobcat and dependencies clean-up.

## Motivation and Context
We should review our functional tests and make them run faster. They might be more compact.

## Types of changes

## Checklist:

---


integration-tests/sanity-functional/src/main/resources/config/common/webdriver.properties
```diff
@@ -0,0 +1,48 @@
...
+
+webdriver.defaultTimeout=10
+page.title.timeout=30
+thread.count=3
+
+
+# +++ Chrome +++
+#webdriver.chrome.driver property should point to Chrome webdriver executable.
+webdriver.type=chrome
+webdriver.chrome.driver=D:/Selenium/chromedriver.exe
```
Skejven: When this path is required. Will this work on all instances?

wiiitek: You are right: this is an arbitrary value. If chromedriver is not found then there is a stacktrace with following error:

```
2) Error in custom provider, java.lang.IllegalStateException: The driver executable does not exist: D:\Selenium\chromedriver.exe
```

If this config value is commented then we have following error:

```
2) Error in custom provider, java.lang.IllegalStateException: The path to the driver executable must be set by the webdriver.chrome.driver system property;
```

What should we do here? Currently on master we have: 

* [integration-tests/sanity-functional/src/main/config/webdriver.properties#L23](https://github.com/Cognifide/aet/blob/master/integration-tests/sanity-functional/src/main/config/webdriver.properties#L23)

Skejven: Maybe we should at least document the necessary change on the path to the chromedriver in the comment on in the [it-tests README](https://github.com/Cognifide/aet/tree/master/integration-tests#sanity-functional) ?

wiiitek: I have update the README

PR #269, opened 2018-06-29
Adding new feature - advanced screen comparison.

aux-lux:
## Description
Script is doing selective comparison of two images to provide precise comparison.

## Motivation and Context
It's way better then recent, obviously.

---

### DIFF
--- a/report/src/main/webapp/assets/js/newComparison/newComparison.js
+++ b/report/src/main/webapp/assets/js/newComparison/newComparison.js
@@ -1,0 +1,698 @@
+window.initNewComparison = function (window, document, undefined) { return function(event) {
+  var debug = false;
+  var tab = document.querySelector('.test-tabs > .tab-content > .tab-pane.ng-scope.active');
+  var wrapper = tab.querySelector('.page-main .layout-compare');
+  var initedClass = 'beta-active';
+  var arrangeTimeout;
+
+  debug && console.log('tab: ', tab);
+  debug && console.log('wrapper: ', wrapper);
+
+  function prepareMarkup(wrapper) {
+    wrapper.innerHTML = '<div class="difs"></div><div class="customMasks"><div class="customMask maskAtop"></div><div class="customMask maskAbot"></div><div class="customMask maskBtop"></div><div class="customMask maskBbot"></div><div class="customLabel labelA"><div class="ruller"></div><div class="text"></div></div><div class="customLabel labelB"><div class="ruller"></div><div class="text"></div></div></div><div class="canvas-wrapper"></div>';
+  }
+
+	if (!wrapper) {
+    console.warn('Can\'t find wrapper');
+		return;
+  } else if (wrapper.classList.contains(initedClass)) {
+    console.warn('Beta already inited');
+    return;
+	} else {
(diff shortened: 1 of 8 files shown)


report/src/main/webapp/assets/js/newComparison/newComparison.js
```diff
@@ -0,0 +1,698 @@
...
+	}
+
+  function simplifyImage(imageData) {
+    var imageWidth = imageData.width;
+    var offset = 0;
+    var data = imageData.data;
+    var simplifiedList = [];
+    var previousLine;
+
+    for (var y = 0; y < data.length / imageWidth / 4; y ++) {
```
zofoklecja: I think `data.length / (imageWidth * 4)` would look clearer

Also - why 4? Magic variable? ✨
I can see it being used throughout the function. Consider putting it in var with descriptive name.

report/src/main/webapp/assets/js/newComparison/newComparison.js
```diff
@@ -0,0 +1,698 @@
+window.initNewComparison = function (window, document, undefined) { return function(event) {
```
zofoklecja: Indents seem to be all over the place in this file. Please fix them for readability.

report/src/main/webapp/assets/js/newComparison/newComparison.js
```diff
@@ -0,0 +1,698 @@
...
+					newDifferenceContextB.putImageData(imageData, 0, 0, 0, 0, imgSizeA.width, heightB);
+				}
+
+				difsA.push(newDifferenceA);
+				difsB.push(newDifferenceB);
+			}
+		}
+	}
+
+	function convertDifferenceData(data1, data2) {
```
zofoklecja: Numbers in variable names usually make for unreadable code.

report/src/main/webapp/assets/js/newComparison/newComparison.js
```diff
@@ -0,0 +1,698 @@
...
+		}
+	});
+
+	function getColor() {
+		var color = colors[colorIndex];
+		colorIndex = (colorIndex + 1) % colors.length;
+		return color;
+	}
+
+	function processLines() {
```
zofoklecja: This is one monster of a function. Any way to split it into more readable chunks or simplify the logic you use?

report/src/main/webapp/assets/js/newComparison/newComparison.js
aux-lux: It's because of way data from image is provided. It reads all pixels and merges it into single array:
[r0,g0,b0,a0,r1,g1,b1,a1,...] r0 = red value from first pixel, r1 = red value from second pixel etc.
It can be named tho. I'll do it.

report/src/main/webapp/assets/js/newComparison/newComparison.js
aux-lux: Can be tricky :) I'll try ;)

Skejven:
All fixes done in #297 , thank wyou @aux-lux and @m-suchorski

PR #289, opened 2018-07-18
Fixed #70 - user now stays in the same tab while navigating around th…

### LINKED ISSUE #70: Navigating between pages should remain user on the same tab
Let's say you want to check W3C errors on each page in your suite. When you navigate between pages, you are getting pulled back to the first tab. 

I would expect to remain on the same tab/aspect of test when navigating through pages.

m-suchorski:
User now stays in the same tab while navigating around tests tree.

## Description
Moving around the tree by using *[ ]* or arrows now automatically opens up url view instead of test view. User also stays on the same tab when moving between urls. If the same type of tab isn't found on the next/previous test url then first tab is selected.

## Motivation and Context
Fix for issue:  #70 - *Navigating between pages should remain user on the same tab*

## Types of changes

## Checklist:

---

### DIFF
--- a/report/src/main/webapp/app/components/keyboardShortcuts.directive.js
+++ b/report/src/main/webapp/app/components/keyboardShortcuts.directive.js
@@ -142,7 +142,7 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
             if (!_.isEmpty($(suiteContainer).nextAll(
                     '.aside-report:not(.is-hidden)').first().find(
                     '.test-name'))) {
-              currentTabLabel = $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
+              currentTabLabel = getCurrentTabLabel();
               $nextElement = suiteContainer.nextAll(
                     '.aside-report:not(.is-hidden)').first();
               handleStyling(testContainer, currentItem, $nextElement);
@@ -150,7 +150,7 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
               userSettingsService.setLastTab(currentTabLabel);
             }
           } else {
-            currentTabLabel = $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
+            currentTabLabel = getCurrentTabLabel();
             userSettingsService.setLastTab(currentTabLabel);
             nextUrl.click();
             scrollTo(nextUrl);
@@ -187,7 +187,7 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
             if (!_.isEmpty($(suiteContainer).prevAll(
                   '.aside-report:not(.is-hidden)').first()
                   .find('.test-name'))) {
-              currentTabLabel = $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
+              currentTabLabel = getCurrentTabLabel();
               $previousElement = suiteContainer.prevAll(
                 '.aside-report:not(.is-hidden)').first();
               handleStyling(testContainer, currentItem, $previousElement);
@@ -195,7 +195,7 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
               userSettingsService.setLastTab(currentTabLabel);
             }
           } else {
-            currentTabLabel = $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
+            currentTabLabel = getCurrentTabLabel();
             userSettingsService.setLastTab(currentTabLabel);
             scrollTo(previousTest);
             previousTest.click();
@@ -240,6 +240,10 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
           .find('.test-url')
           .last()
           .click();
+      }
+
+      function getCurrentTabLabel() {
+        return $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
       }
 
       function clickOnTab() {
--- a/report/src/main/webapp/app/components/keyboardShortcuts.directive.js
+++ b/report/src/main/webapp/app/components/keyboardShortcuts.directive.js
@@ -177,7 +177,6 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
             currentLocation = window.location.hash,
             $previousElement,
             currentTab;
-
         if (!(ifRootPage(currentLocation, '/url/') || ifRootPage(
                 currentLocation, '/test/') || ifRootPage(currentLocation,
                 '/report/'))) {
--- a/report/src/main/webapp/app/components/keyboardShortcuts.directive.js
+++ b/report/src/main/webapp/app/components/keyboardShortcuts.directive.js
@@ -125,7 +125,7 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
             currentLocation = window.location.hash,
             $nextElement,
             $firstElementInTest,
-            currentTab;
+            currentTabLabel;
 
         if (!(ifRootPage(currentLocation, '/url/') || ifRootPage(
                 currentLocation, '/test/') || ifRootPage(currentLocation,
@@ -142,24 +142,19 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
             if (!_.isEmpty($(suiteContainer).nextAll(
                     '.aside-report:not(.is-hidden)').first().find(
                     '.test-name'))) {
-              currentTab = $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
-              testContainer.removeClass('is-active');
-              currentItem.removeClass('is-active');
+                      currentTabLabel = $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
               $nextElement = suiteContainer.nextAll(
-                  '.aside-report:not(.is-hidden)').first();
-              $nextElement.addClass('is-expanded');
-              $nextElement.children().first().addClass('is-active');
-              scrollTo($nextElement.find('.is-active'));
+                    '.aside-report:not(.is-hidden)').first();
+              handleStyling(testContainer, currentItem, $nextElement);
               toggleNextTestItem(suiteContainer);
-              userSettingsService.setLastTab(currentTab);
+              userSettingsService.setLastTab(currentTabLabel);
             }
           } else {
-            currentTab = $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
-            userSettingsService.setLastTab(currentTab);
+            currentTabLabel = $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
+            userSettingsService.setLastTab(currentTabLabel);
             nextUrl.click();
             scrollTo(nextUrl);
             $(testUrlSelector).not(nextUrl).removeClass('is-active');
-            clickOnTab();
           }
         } else {
           currentTest = findCurrentTest(currentLocation.split('/').pop());
@@ -176,7 +171,8 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
         var previousTest,
             currentLocation = window.location.hash,
             $previousElement,
-            currentTab;
+            currentTabLabel;
+
         if (!(ifRootPage(currentLocation, '/url/') || ifRootPage(
                 currentLocation, '/test/') || ifRootPage(currentLocation,
                 '/report/'))) {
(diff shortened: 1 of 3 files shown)


report/src/main/webapp/app/components/keyboardShortcuts.directive.js
```diff
@@ -141,36 +142,41 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
             if (!_.isEmpty($(suiteContainer).nextAll(
                     '.aside-report:not(.is-hidden)').first().find(
                     '.test-name'))) {
+              currentTab = $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
```
adamkosmala: is it possible to have more than one `.nav-item` element with class `active`? if not, why not going without filtering?

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
```diff
@@ -141,36 +142,41 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
...
                   '.aside-report:not(.is-hidden)').first();
               $nextElement.addClass('is-expanded');
               $nextElement.children().first().addClass('is-active');
               scrollTo($nextElement.find('.is-active'));
+              toggleNextTestItem(suiteContainer);
+              userSettingsService.setLastTab(currentTab);
             }
-
           } else {
+            currentTab = $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
```
oliwerin: Please add a `$` prefix to the variable name to indicate it holds a jQuery object

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
adamkosmala: after discussion I'm ok with keeping it as it is

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
```diff
@@ -183,14 +189,26 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
...
           if (_.isEmpty(previousTest)) {
-            $previousElement = suiteContainer.find('.test-name');
-            scrollTo($previousElement);
-            $previousElement.click();
-            currentItem.parents('.url-name').removeClass('is-active');
-            suiteContainer.addClass('is-expanded');
+            if (!_.isEmpty($(suiteContainer).prevAll(
+                  '.aside-report:not(.is-hidden)').first()
+                  .find('.test-name'))) {
+              currentTab = $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
```
oliwerin: Here you have `active` and in the lines below `is-active` - can we make it consistent?

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
m-suchorski: Even though it's just plain text at the end ('screenlayout' for example)?

oliwerin: Ok, didn't notice - no need to worry about it then.

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
m-suchorski: .active is a class for tabs while .is-active is a class for sidebar menu, I didn't modify any html/css structure. I can ask @Skejven what he thinks about it and then maybe change it.

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
oliwerin: The name is very misleading though - rename it to `currentTabLabel` or something else more accurate

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
```diff
@@ -208,13 +226,54 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
...
+          .first()
+          .find('.test-url')
+          .last()
+          .click();
+      }
+
+      function clickOnTab() {
+        var currentTab = userSettingsService.getLastTab();
+        var nextTestTabs = $('.nav-tabs').children();
+        for(var i=0;i <nextTestTabs.length; i++) {
```
oliwerin: Why you're not using jQuery `each` for keeping it simpler and more consistent?

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
```diff
@@ -208,13 +226,54 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
...
+          '.aside-report:not(.is-hidden)')
+          .first()
+          .find('.test-url')
+          .last()
+          .click();
+      }
+
+      function clickOnTab() {
+        var currentTab = userSettingsService.getLastTab();
+        var nextTestTabs = $('.nav-tabs').children();
```
oliwerin: Both lines - 248-249 - consider adding a `$` suffix if they store jQuery collections

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
```diff
@@ -208,13 +226,54 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
...
+        for(var i=0;i <nextTestTabs.length; i++) {
+          if($(nextTestTabs[i]).text().replace(/\s/g, '') === currentTab) {
+            $(nextTestTabs[i]).find('a').click();
+            currentTab = null;
+          }
+        }
+      }
+
+      var mutationObserver = new MutationObserver(callback);
```
oliwerin: Can you shortly explain why it's needed with config set to `true` for each property?

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
m-suchorski: Here again currentTab is just plain text, will change that to currentTabLabel.
As to next line I will change that.

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
m-suchorski: I've used this: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserverInit 
I've tried few configs and this one has been working properly for me.

oliwerin: Ok, but why do you need MutationObserver, whatever its config is?

m-suchorski: It looks for changes in the DOM before it is displayed to the user.

Here, if I didn't use MutationObserver my screen would flick every time I changed tree URL, because  selecting the tab is done by using .click(); 

With MutationObserver I can atually click something before it is displayed (but it already exists in the DOM), both the delay and the flickering is gone.

I've tried using timeout - but it is based on how many tests are performed, the more tests, the longer the timeout should be (and it varies from 10 to 50ms +). Then I've tried with window.onhashchange but it didn't change anything. I just needed something that would detect a change before user sees anything.

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
Skejven: @oliwerin what is the problem with existing those two classes?
If you recommend to align them, please do so in a separate pull request.

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
```diff
@@ -183,14 +189,26 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
               '.url-name:not(.is-hidden)').filter(':first').find(
               testUrlSelector);
           if (_.isEmpty(previousTest)) {
-            $previousElement = suiteContainer.find('.test-name');
-            scrollTo($previousElement);
-            $previousElement.click();
-            currentItem.parents('.url-name').removeClass('is-active');
-            suiteContainer.addClass('is-expanded');
+            if (!_.isEmpty($(suiteContainer).prevAll(
```
Skejven: Maybe it would be worth to put following 20 lines into a separate method? `traverseTreeUp` is a little to long now.

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
Skejven: Could you also fix formatting? Thanks.

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
```diff
@@ -208,13 +226,54 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
...
+            currentTab = null;
+          }
+        }
+      }
+
+      var mutationObserver = new MutationObserver(callback);
+
+      function callback(mutList) {
+        var finished = false;
+        mutList.forEach(function(mut) {
```
Skejven: How about using `Array.prototype.find()` insead of `forEach`? You wouldn't have to do this nasty `finished` guard and finish `forEach` loop (better performance).

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
oliwerin: It's worth to keep in mind that MutationObserver with such an extensive config might be performance heavy. It also sounds to me that it's used to fix a problem that should not occur - I don't think that screen should flick every time you change the URL. Maybe it would be better to spend more time on it and fix it properly.
I would suggest at least adding a short comment explaining why MutationObserver is used so future developers working on it have a clear understanding.

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
oliwerin: @Skejven It's not a huge issue but it's a class representing certain state and ideally there should be only one - for better consistency, readability and maintainability.

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
```diff
@@ -208,13 +216,64 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
...
+          '.aside-report:not(.is-hidden)')
+          .first()
+          .find('.test-url')
+          .last()
+          .click();
+      }
+
+      function clickOnTab() {
+        var currentTabLabel = userSettingsService.getLastTab();
+        var $nextTestTabs = $('.nav-tabs').children();
```
oliwerin: In some places you use multiple `var` notation (here) and in others one `var` and variable declaration separated out by a comma. I'd be good to stick with just one of them.

tkaik:
Please update Changelog file https://github.com/Cognifide/aet/blob/master/CONTRIBUTING.md#changelog

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
```diff
@@ -141,36 +142,36 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
             if (!_.isEmpty($(suiteContainer).nextAll(
                     '.aside-report:not(.is-hidden)').first().find(
                     '.test-name'))) {
-              testContainer.removeClass('is-active');
-              currentItem.removeClass('is-active');
-              toggleNextTest(suiteContainer);
+                      currentTabLabel = $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
```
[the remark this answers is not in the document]
m-suchorski: Right, didn't notice. Fixed now.

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
m-suchorski: Changed, but I still got some stuff to do (with MutationObserver) so I will commit tomorrow.

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
```diff
@@ -141,36 +142,36 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
             if (!_.isEmpty($(suiteContainer).nextAll(
                     '.aside-report:not(.is-hidden)').first().find(
                     '.test-name'))) {
-              testContainer.removeClass('is-active');
-              currentItem.removeClass('is-active');
-              toggleNextTest(suiteContainer);
+              currentTabLabel = $('.nav-tabs > .nav-item').filter('.active').text().replace(/\s/g, '');
```
tkaik: This line is repeated across the file few times - can it be moved to a seperate function, e.g. `getCurrentTabLabel`?

m-suchorski: Fixed.

PR #301, opened 2018-08-02
New virtualization tool set (and new aet-cookbook version)

wiiitek:
Upgrade for our virtualization tools to:

* [VirtualBox 5.2.18](https://www.virtualbox.org/wiki/Downloads)
* [Vagrant 2.1.2](https://releases.hashicorp.com/vagrant/)
* [ChefDK 3.1.0](https://downloads.chef.io/chefdk/stable)

## Description
Updates made fr cookbook are on separate branch. This branch is specified in `Berksfile`.
When we are ready to merge (tests on different environments for new AET 10 minut setup) we would need to release new aet-cookbook version.

## Motivation and Context
New Virtualbox and Chef will allow us to make progress into Windows environments more easily.
And we also don't want to use some old, non-supported versions ([VirtualBox 5.1.x is no longer supported!](https://www.virtualbox.org/wiki/Download_Old_Builds_5_1)).

## Types of changes

## Checklist:

---


vagrant/Vagrantfile
```diff
@@ -96,7 +96,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       }
     }
     chef.run_list = [
-      "recipe[aet::_develop]",
+#      "recipe[aet::_develop]",
```
tkaik: Thank you for this PR @wiiitek :)
I think we should use `aet::_develop` by default on Vagrant - it creates `developer` user which allows us to easily upload AET to Vagrant with maven using `developer/developer` credentials, as instructed in [AET in 10 minutes](https://github.com/Cognifide/aet/wiki/AETIn10Minutes#upload-vagrant-prequisities)

wiiitek: You're right! This change was to test the use case described in https://github.com/Cognifide/aet-cookbook/pull/30#discussion_r212208661. We should probably do a release for the aet-cookbook (version 5.0.0). What do you think?

PR #306, opened 2018-08-07
Update osgi annotations in runner

plutasnyy:
I updated osgi annotations in runner

## Motivation and Context
Before, annotations was marked as deprecated.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the [code style](https://github.com/Cognifide/aet/blob/master/CONTRIBUTING.md#coding-conventions) of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

---
I hereby agree to the terms of the AET Contributor License Agreement.

PR #309, opened 2018-08-07
Updates for Pull request #306

wiiitek:
I have some additional updates for https://github.com/Cognifide/aet/pull/306

## Motivation and Context
I think it is better for code quality when we don't use Powermock :)
Also let's remove SCR annotations from poms.

## Types of changes

## Checklist:

---

### DIFF
--- a/core/runner/pom.xml
+++ b/core/runner/pom.xml
@@ -110,14 +110,6 @@
       <groupId>com.googlecode.zohhak</groupId>
       <artifactId>zohhak</artifactId>
     </dependency>
-    <dependency>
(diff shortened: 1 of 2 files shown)


core/runner/pom.xml
```diff
@@ -141,6 +129,17 @@
           </execution>
         </executions>
       </plugin>
+      <!-- overwrites configuration from parent pom to disable SCR plugin -->
```
Skejven: We probably should add `<!-- ToDo: remove me when all felix dependencies are removed -->` here.

PR #310, opened 2018-08-08
Update osgi annotations in rest endpoint

plutasnyy:
I updated osgi annotations in rest endpoint

## Motivation and Context

Before, annotations was marked as deprecated

## Types of changes

## Checklist:

---


plutasnyy:
My bad, I forgot add files to git repository ;)

rest-endpoint/src/main/java/com/cognifide/aet/rest/MetadataServlet.java
```diff
@@ -35,19 +35,18 @@
...
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
+import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.http.HttpService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@Service
-@Component(label = "MetadataServlet", description = "Returns Suite Metadata", immediate = true)
+@Component(service = BasicDataServlet.class, immediate = true)
```
Skejven: Could you check if `service = HttpServlet.class` would be enough here and in other `BasicDataServlet` extensions please? We only want this class to be registered as a servlet. It won't be referenced anywhere.

test-executor/pom.xml
```diff
@@ -123,6 +123,7 @@
 
   <build>
     <plugins>
+      <!-- ToDo: remove me when all felix dependencies are removed -->
```
Skejven: Please clear also dependencies:
```
   <dependency>
      <groupId>org.apache.sling</groupId>
      <artifactId>org.apache.sling.commons.osgi</artifactId>
    </dependency>
```
and
```
    <dependency>
      <groupId>org.apache.felix</groupId>
      <artifactId>org.apache.felix.scr.annotations</artifactId>
    </dependency>
```

plutasnyy: I edited wrong file, now should be correct

PR #314, opened 2018-08-08
updated osgi annotatation in jobs

plutasnyy:
Migrate Felix annotations to OSGi in Jobs module

## Types of changes

## Checklist:

---

### DIFF
--- a/core/jobs/pom.xml
+++ b/core/jobs/pom.xml
@@ -36,6 +36,18 @@
   <dependencies>
     <!-- project -->
     <dependency>
+      <groupId>org.osgi</groupId>
+      <artifactId>org.osgi.service.component.annotations</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.osgi</groupId>
+      <artifactId>org.osgi.annotation</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.osgi</groupId>
+      <artifactId>org.osgi.service.metatype.annotations</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.cognifide.aet</groupId>
       <artifactId>communication-api</artifactId>
     </dependency>
--- a/core/jobs/pom.xml
+++ b/core/jobs/pom.xml
@@ -91,10 +91,6 @@
       <artifactId>diffutils</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.apache.sling</groupId>
-      <artifactId>org.apache.sling.commons.osgi</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>fluent-hc</artifactId>
     </dependency>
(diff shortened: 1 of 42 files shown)


core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/executejavascript/ExternalSnippetHttpClient.java
```diff
@@ -37,42 +32,34 @@
...
 
-@Component
-@Service(ExternalSnippetHttpClient.class)
+@Component(service = ExternalSnippetHttpClient.class)
+@Designate(ocd = ExternalSnippetHttpClientConf.class)
 public class ExternalSnippetHttpClient {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ExternalSnippetHttpClient.class);
 
   private static final long DEFAULT_CONNECTION_TTL = 60L;
```
Skejven: Those values were moved to `ExternalSnippetHttpClientConf` so you probably can remove it from here.

core/jobs/pom.xml
```diff
@@ -35,6 +35,18 @@
 
   <dependencies>
     <!-- project -->
+    <dependency>
```
Skejven: You may get rid of
```xml
    <dependency>
      <groupId>org.apache.felix</groupId>
      <artifactId>org.apache.felix.scr.annotations</artifactId>
    </dependency>
```
and 
```xml
    <dependency>
      <groupId>org.apache.sling</groupId>
      <artifactId>org.apache.sling.commons.osgi</artifactId>
    </dependency>
```

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/executejavascript/ExternalSnippetHttpClient.java
plutasnyy: Removed

core/jobs/pom.xml
plutasnyy: Done

core/jobs/src/main/java/com/cognifide/aet/job/common/collectors/source/configuration/SourceCollectorFactoryConf.java
```diff
@@ -0,0 +1,35 @@
...
+
+  String SOURCE_COLLECTOR_TIMEOUT_VALUE = "sourceCollectorTimeoutValue";
+
+  int DEFAULT_TIMEOUT = 20000;
+
+  @AttributeDefinition(
+      name = SOURCE_COLLECTOR_TIMEOUT_VALUE,
+      description =  "Timeout value for source collector [ms]",
+      type = AttributeType.INTEGER)
+  int timeoutValue() default  DEFAULT_TIMEOUT;
```
tkaik: With new [@AttributeDefinition](https://osgi.org/javadoc/r6/cmpn/org/osgi/service/metatype/annotations/AttributeDefinition.html) annotation, the `name` property is just a human-readable name (i.e. label for the property) - the property name itself is identified by the method name (`timeoutValue`). So now the method names should be equal to the value of `name` property on the old `@Property` annotations - otherwise we'll need to change property names in all karaf configuration files for this component.

PR #328, opened 2018-08-16
Suite's history

m-suchorski:
User can now view and go to any version of that suite.

## Description

User is now able to go trough all versions of the suite by clicking on the timestamp. 
Suite's versions are displayed after clicking the current version on the top toolbar.

## Types of changes

## Checklist:

---


api/datastorage-api/src/main/java/com/cognifide/aet/vs/MetadataDAO.java
```diff
@@ -47,6 +47,14 @@
    */
   Suite getSuite(DBKey dbKey, String correlationId) throws StorageException;
 
+  /**
+   * @param dbKey - key with project and company name
+   * @param name - name of suite
+   * @param version Suite object found by given criteria or null.
```
tkaik: Please fix this javadoc - move `Suite object found by given criteria or null.` to `@return` and add description for `version` param

core/datastorage/src/main/java/com/cognifide/aet/vs/metadata/MetadataDAOMongoDBImpl.java
```diff
@@ -143,6 +157,21 @@ public Suite apply(Document result) {
     }).toList();
   }
 
+  public List<String> listSuiteVersions(DBKey dbKey, String name) throws StorageException{
```
tkaik: This method should return list of pairs: `{ version: "..", correlationId: "..}` - we can't assume that oldest suite will have `version` equal to `1` because Cleaner might have deleted it

report/src/main/webapp/app/layout/toolbar/toolbarTop.controller.js
```diff
@@ -17,36 +17,95 @@
...
+            '&correlationId=' + vm.suiteStatistics.patternCorrelationId
         };
       }
     }
 
+    function buildApiPath($allParametersList) {
+      return 'http://aet-vagrant' + '/api' + '/history' + '?' +
+        $allParametersList[0] + '&' +
+        $allParametersList[1] + '&' +
+        $allParametersList[2];
```
PiteroS8: The path is wrong when correlationId is used.
Example: `/report.html?company=xxx&project=xxx&correlationId=xxx`

core/datastorage/src/main/java/com/cognifide/aet/vs/metadata/MetadataDAOMongoDBImpl.java
```diff
@@ -143,6 +157,21 @@ public Suite apply(Document result) {
...
+    MongoCollection<Document> metadata = getMetadataCollection(dbKey);
+    LOGGER.debug("Fetching all versions for suite: `{}` , company: `{}`, project: `{}`.", name, dbKey.getCompany(),
+        dbKey.getProject());
+
+    final FindIterable<Document> found = metadata
+        .find(Filters.eq(SUITE_PARAM_NAME, name))
+        .sort(Sorts.descending(SUITE_VERSION_PARAM_NAME));
+
+    return FluentIterable.from(found).transform(new Function<Document, String>() {
```
tkaik: You may use Java 8 features instead, e.g.:
```
StreamSupport.stream(found.spliterator(), false)
        .map(result -> result.getString("correlationId"))
        .collect(Collectors.toList());
```

core/datastorage/src/main/java/com/cognifide/aet/vs/metadata/MetadataDAOMongoDBImpl.java
```diff
@@ -135,12 +152,26 @@ public Suite getLatestRun(DBKey dbKey, String name) throws StorageException {
...
-      public Suite apply(Document result) {
-        return new DocumentConverter(result).toSuite();
-      }
-    }).toList();
+    return StreamSupport.stream(found.spliterator(), false)
+        .map((Function<Document, Suite>) document -> new DocumentConverter(document).toSuite())
+        .collect(Collectors.toList());
+  }
+
+  public List<List<String>> listSuiteVersions(DBKey dbKey, String name) throws StorageException {
```
tkaik: This method should return `List<Map<String, String>>` or `List<SuiteVersion>` (where `SuiteVersion` is a new class with 2 fields - `version` and `correlationId`. Then in `HistoryServlet` you just need to call `gson.toJson` instead of the whole `JsonArray`/`JsonObject` building

report/src/main/webapp/app/layout/toolbar/toolbarTop.controller.js
```diff
@@ -17,36 +17,124 @@
...
+    var suiteHeaders;
+    getSuiteHistory(suiteHeaders, $rootScope, $http);
     updateToolbar();
 
     /***************************************
      ***********  Private methods  *********
      ***************************************/
 
     function updateToolbar() {
+      document.getElementsByClassName('suite-history-container')[0].style.display = 'none';
```
Skejven: You can use jQuery here.

report/src/main/webapp/app/layout/toolbar/toolbarTop.controller.js
```diff
@@ -17,36 +17,124 @@
...
-          '&project=' + vm.suiteStatistics.project +
-          '&correlationId=' + vm.suiteStatistics.patternCorrelationId
+            '&project=' + vm.suiteStatistics.project +
+            '&correlationId=' + vm.suiteStatistics.patternCorrelationId
         };
       }
     }
 
+    function buildApiPath($allParametersList) {
+      return 'http://aet-vagrant' + '/api' + '/history' + '?' +
```
Skejven: Please use AET endpoint address defined in the `endpointConfigurationService`.

report/src/main/webapp/app/layout/toolbar/toolbarTop.controller.js
```diff
@@ -17,36 +17,124 @@
...
+        $allParametersList[0] + '&' +
+        $allParametersList[1] + '&' +
+        $allParametersList[2];
+    }
+
+    function getSuiteHistory(suiteHeaders, $rootScope, $http) {
```
Skejven: This kind of logic should be placed into angular service not in the controller.
E.g. look at `metadataEndpointService`.

report/src/main/webapp/app/layout/toolbar/toolbarTop.view.html
```diff
@@ -83,5 +86,31 @@
     <div class="toolbar-block fontawesome-link" data-toggle="modal" data-target="#helpModal">
       <i class="fas fa-question-circle fa-lg"></i>
     </div>
+    <div class="suite-history-container">
```
Skejven: Maybe it would be good idea to put that into separate modal? Like e.g. `noteModal` (separate controller and view).

rest-endpoint/src/main/java/com/cognifide/aet/rest/MetadataServlet.java
```diff
@@ -75,7 +76,11 @@ protected void process(DBKey dbKey, HttpServletRequest req, HttpServletResponse
       if (isValidCorrelationId(correlationId)) {
         suite = metadataDAO.getSuite(dbKey, correlationId);
       } else if (isValidName(suiteName)) {
-        suite = metadataDAO.getLatestRun(dbKey, suiteName);
+        if(isValidName(suiteVersion)){
```
Skejven: Shouldn't there be method `isValidVersion` here?

report/src/main/webapp/app/layout/toolbar/toolbarTop.view.html
m-suchorski: Done.

report/src/main/webapp/app/layout/toolbar/toolbarTop.controller.js
m-suchorski: Done.

report/src/main/webapp/app/layout/toolbar/toolbarTop.controller.js
m-suchorski: Done.

report/src/main/webapp/app/layout/toolbar/toolbarTop.controller.js
m-suchorski: Done.

documentation/src/main/wiki/RestAPI.md
```diff
@@ -43,6 +51,14 @@ Base api path:
 * **Example**: http://aet.example.com/api/metadata `[raw json in post body]`
 * **Description**: This method increments version number before saving to DB and returns updated suite object in json format. Returns status 409 if given suite is locked.
 
+--------
+##### Get correlationIds and versions by suite name
+* **URL**: `/api/history`
+* **HTTP Method**: GET
+* **Parameters**: `company`, `project`, `suite`
+* **Example**: http://aet.example.com/api/history??company=cognifide&project=example&suite=mysimplesuite
```
tkaik: Please fix typo - double `?`

report/src/main/webapp/app/services/history.service.js
```diff
@@ -0,0 +1,136 @@
...
+define(['angularAMD', 'endpointConfiguration'], function (angularAMD) {
+  'use strict';
+  angularAMD.factory('historyService', historyService);
+
+  /**
+   * Service responsible for fetching suite's history
+   */
+  function historyService($rootScope, $http, endpointConfiguration) {
+    var suiteHeaders;
+    $rootScope.endpointUrl = endpointConfiguration.getEndpoint().getUrl;
```
Skejven: Is that necessary to bind this value to the `$rootScope` variable?
Maybe you can use local variable instead?

documentation/src/main/wiki/SuiteReportFeatures.md
```diff
@@ -45,6 +45,18 @@ Moreover, every URL has information to which test group it belongs.
 
 ![Find example](assets/suiteReport/find-in-sidepanel-example.png)
 
+#### Browsing suite's history
+It is possible to view current suite's history by clicking current suite's name and version in the top toolbar. User can also navigate through versions by clicking next/previous arrow next to suite's version. 
+
+![Suite history](assets/suiteReport/suite-history.png)
+
+Clicking the suite's name in the top toolbar will show a popup in which the user can see every suite's version that is still available in the database. User can also directly go to the newest version of the suite by clicking
```
tkaik: It looks like unfinished sentence :)

rest-endpoint/src/main/java/com/cognifide/aet/rest/MetadataServlet.java
```diff
@@ -77,7 +78,11 @@ protected void process(DBKey dbKey, HttpServletRequest req, HttpServletResponse
       if (isValidCorrelationId(correlationId)) {
         suite = metadataDAO.getSuite(dbKey, correlationId);
       } else if (isValidName(suiteName)) {
-        suite = metadataDAO.getLatestRun(dbKey, suiteName);
+        if(isValidVersion(suiteVersion)){
+          suite = metadataDAO.getSuite(dbKey, suiteName, suiteVersion);
+        }else {
```
Skejven: Please fix formatting here

report/src/main/webapp/app/services/history.service.js
m-suchorski: This value isn't used anywhere in the service file - it is used only in the view file so it has to be put in the `$rootScope`, otherwise it wouldn't be possible to use it in the `historyModal.view.html` file.

documentation/src/main/wiki/SuiteReportFeatures.md
m-suchorski: Right, I don't know what happened there. Fixing now.

report/src/main/webapp/app/services/history.service.js
```diff
@@ -0,0 +1,136 @@
...
+
+    function getSuiteHistory(suiteHeaders, $rootScope, $http, historyCallback) {
+      $rootScope.data = 'test';
+      var cUrl = new URL(window.location.href);
+      var company = cUrl.searchParams.get('company');
+      var project = cUrl.searchParams.get('project');
+      var suite = cUrl.searchParams.get('suite');
+      if (suite === null) {
+        var correlationId = cUrl.searchParams.get('correlationId');
+        suite = correlationId.split('-')[2];
```
tkaik: It is possible that `company` or `project` or `suite` values contain `-` character - then this line will not work correctly.
I think it can be done like this:
```
var cp = company + '-' + project + '-';
var suiteNameWithTimestamp = correlationId.substr(a.indexOf(cp) + cp.length)
// e.g. now: suiteNameWithTimestamp == 'suite-name-123123123` - we need to cut timestamp from here
```

report/src/main/webapp/app/services/history.service.js
```diff
@@ -0,0 +1,136 @@
...
+        }
+      }).then(function (response) {
+        $rootScope.data = response.data;
+        suiteHeaders = response.data;
+        for (var i = 0; i < suiteHeaders.length; i++) {
+          var version = suiteHeaders[i].version;
+          var company = suiteHeaders[i].correlationId.split('-')[0];
+          var project = suiteHeaders[i].correlationId.split('-')[1];
+          var suite = suiteHeaders[i].correlationId.split('-')[2];
+          var correlationId = suiteHeaders[i].correlationId.split('-')[3];
```
tkaik: I think this you're parsing timestamp from the correlationID here, so this var should be renamed to e.g. `timestamp`
And as above - we should handle cases where company/project/suite have '-' characters

report/src/main/webapp/app/services/history.service.js
Skejven: Where exactly in `historyModal.view.html` is used `endpointUrl`?

report/src/main/webapp/app/services/history.service.js
```diff
@@ -0,0 +1,136 @@
...
+        }
+      }).then(function (response) {
+        $rootScope.data = response.data;
+        suiteHeaders = response.data;
+        for (var i = 0; i < suiteHeaders.length; i++) {
+          var version = suiteHeaders[i].version;
+          var company = company;
+          var project = project;
+          var suite = suite;
+          var timestamp = suiteHeaders[i].correlationId.split('-')[suiteHeaders[i].correlationId.split('-').length - 1];
```
tkaik: This is a bit hard to read now, can we make `suiteHeaders[i].correlationId.split('-')` a variable? e.g.:
```
var correlationIdParts = suiteHeaders[i].correlationId.split('-');
var timestamp = correlationIdParts[correlationIdParts.length - 1]
```

report/src/main/webapp/app/services/history.service.js
```diff
@@ -0,0 +1,136 @@
...
+          suiteHeaders[i] = {
+            company: company,
+            project: project,
+            suite: suite,
+            version: version,
+            correlationId: timestamp,
```
tkaik: `correlationId: timestamp` is a bit confusing, please change to `timestamp: timestamp` (because it's not correlatoinId here)

PR #331, opened 2018-08-22
Feature/support env variable in mongo config

Skejven:
## Description
MongoDB address was configurable only from the OSGi confi level. When using Docker, it is not desired to set OSGi configs manually. MongoDB could be one of the resources outside the Docker cluster.

## Motivation and Context
Enable setting `mongoURI` as an environment variable. It will override any OSGi settings if provided.
This feature will be documented in the future in the Docker support section.
The priority of setting this value is:

- OSGi configuration
- env variable `MONGODB_URI`
- default configuration value

## Types of changes

## Checklist:

---


core/datastorage/src/main/java/com/cognifide/aet/vs/mongodb/MongoDBClient.java
```diff
@@ -103,7 +102,17 @@ public void deactivate() {
 
     return companies;
   }
-  
+
+  protected void setupConfiguration(MongoDBClientConf config) {
+    if (StringUtils.isNoneBlank(config.mongoURI())) {
```
tkaik: You can use `isNotBlank` instead of `isNoneBlank` here :)

core/datastorage/src/main/java/com/cognifide/aet/vs/mongodb/MongoDBClient.java
```diff
@@ -103,7 +102,17 @@ public void deactivate() {
...
+      this.mongoUri = config.mongoURI();
+    } else {
+      this.mongoUri = StringUtils.defaultString(
```
tkaik: What about a case where the ENV varialbe is an empty string? If I'm not mistaken the `StringUtils.defaultString` only checks for null

core/datastorage/src/main/java/com/cognifide/aet/vs/mongodb/MongoDBClient.java
Skejven: Typo - thanks, fixed.

core/datastorage/src/main/java/com/cognifide/aet/vs/mongodb/MongoDBClient.java
```diff
@@ -103,7 +103,15 @@ public void deactivate() {
 
     return companies;
   }
-  
+
+  protected void setupConfiguration(MongoDBClientConf config) {
+    mongoUri = Optional.of(config.mongoURI())
```
tkaik: I think `config.mongoURI()` can return null, don't we need `Optional.ofNullable` here?

Skejven: This is very odd for me, I had `ofNullable` at the beginning but my IDE prompted it as:
2018-08-23_09h02_41
I've tried to find any information if `@ObjectClassDefinition` and OSGi mechanics adds default by itself. Official Java docs: https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.6.2 won't mention explicitly what happens if you don't declare default value (I mean, that there will be `null` returned)...
When I checked it on debug, there actually was `null` - thanks for spotting that!
I will change that. I see that it looks like a good practice to return always some default value and I will set it to empty string in `mongoURI()`.

Issue #317, opened 2018-08-09
Keyboard shortcuts for 'Accept/Revert url/test' buttons


dlewandowski-cognifide:
### Version used

* `AET 2.1.6`

### Context of the issue/feature

There are buttons for accepting/reverting collected patterns in AET reporting App (check attached screens), but it is inconvenient to use them while navigating with keybaord shortcuts bettween different reports, because the user has to click them using mouse cursor. 

***The aim of this task is*** to provide dedicated keyboard shortcuts for accept/revert buttons.

***Solution proposition:***
Introduce following keyboard shortcuts:
***A*** - for accepting suite/test/url
***R*** - for reverting suite/test/url
***ctrl*** - modifier that will define that the specific opened test case should be accepted/reverted (***ctrl + A*** - accepts opened test case ; ***ctrl + R*** - reverts opened test case )

sp2mbn:
Ctrl+A and Ctrl+R are not proper shortcuts, because are natively used in browsers. IMHO Shift key will be better as a modifier.

kaczymuczy:
If someone tries to enter "R" into the filter input then the shortcut will execute or not?

sp2mbn:
I think it shouldn't, but existing shortcuts (e.q. "q" or "e") are executed. If it's not wanted behavior that all existing shortcuts should be fixed.

PR #360, opened 2018-09-18
Keyboard shortcuts for 'Accept/Revert url/test' buttons

sp2mbn:
## Description

- Added shortcuts for Accept/Revert buttons and Shitf key as a modifier for Accept/Revert only selected test case
- Fixed problem of executing shortcuts when cursor is in search input

## Motivation and Context

#317 

## Types of changes

## Checklist:

---


report/src/main/webapp/app/components/keyboardShortcuts.directive.js
```diff
@@ -60,14 +62,16 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
         },
         '221': function () {
           scope.traverseTree('down');
-        }
+        },
+        '65': accept,
```
plutasnyy: Great job!
What do you think, maybe we could remember ASCII codes in variables instead of add comment?

documentation/src/main/wiki/releases/2.0.0/Documentation-2.0.0.md
```diff
@@ -3410,6 +3410,9 @@ It is possible to search tests and URLs by the query. Searched fields are the UR
 * use **[** / **]** to navigate between test's urls
 * press **m** to show/hide layout mask (when available)
 * press **←** / **→** to navigate between url's tabs
+* press **a** for accepting suite/test/url
```
plutasnyy: Please update only `SuiteReportFeatures.md`. I think that it isn't a good idea to update historical documentation ;)

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
sp2mbn: Actually we don't need to use ASCII codes at all and we can use event.key property instead of event.keyCode which is returning single character that identify keyboard buttons. I added new shortcuts using existing convention. If it is necessarily I can quick refactor it.

report/src/main/webapp/app/components/keyboardShortcuts.directive.js
```diff
@@ -86,6 +90,20 @@ define(['angularAMD', 'userSettingsService'], function (angularAMD) {
...
       };
 
+      function clickButton(buttonClass, event) {
+        var $tab = (event.shiftKey) ? $('.tab-content-toolbar') : $('.toolbar-bottom');
+        var $btns = $tab.find('.toolbar-btns');
+        $btns.find(buttonClass).click();
+      }
+
+      function accept(event) {
+        clickButton('.button-blue', event);
```
Skejven: I'm not sure that this is correct assumption, that element with class `button-blue` should be clicked on the accept event (the same for the `button-red`). More buttons also have those classes, e.g.: 
2018-09-20_09h46_37

Maybe buttons for accepting/reverting should have their own identifiers?

sp2mbn: Assumption is that only buttons inside "toolbar-bottom" or "tab-content-toolbar" will be clicked (function clickButton is looking for those classes first and then looking in their child nodes proper buttons for click). "Save all changes" button is inside "sidepanel" so it will not be clicked.

Asia95:
tested, working well ✔️

Issue #357, opened 2018-09-14
Inconsistent screenshot height when height param in resolution modifier is omitted


michalwronowski:
After switching to chromedriver one of the ways to obtain a full page screenshot is to omit the `height` attribute in `resolution` modifier.
After doing so, height of the window is set dynamically to the value of `document.body.scrollHeight` property.
The problem is that the height is read immediately after the width of the window has been set.
This in worst case scenarios can lead to inconsistencies in window height and in result in screenshot height.

Please consider adding configurable delay between setting the window width and querying the window height.

Discussed solutions include:
1. Configurable "cooldown" in milliseconds which would cause the test to sleep for configured time.
2. Configurable "wait-for-height" algorithm, which would query `document.body.scrollHeight` in some defined periods of time to check if the `document.body.scrollHeight` changes. If defined number of samples would match, modifier would use it to set the window height eventually. This solution would require some threshold, either the number of samples or cumulated time of height readings.

Either one of those solutions or both of them could be implemented.

PR #387, opened 2018-10-11
Hotfix/screen max height

Jakub-Izbicki:
Change max allowed screenshot pixels height to 35k.

## Description
After lowering of max allowed screenshot height from 100k to 15k, some projects were unable to test some of their long pages. After testing, this pr changes the value to 35k pixels.

## Types of changes

## Checklist:

---

### DIFF
--- a/core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/resolution/ResolutionModifier.java
+++ b/core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/resolution/ResolutionModifier.java
@@ -41,7 +41,7 @@ public class ResolutionModifier implements CollectorJob {
 
   private static final String JAVASCRIPT_GET_BODY_HEIGHT = "return document.body.scrollHeight";
 
-  private static final int MAX_SIZE = 15000;
+  private static final int MAX_SIZE = 35000;
 
   private static final int INITIAL_HEIGHT = 300;
 
--- a/integration-tests/test-suite/partials/layout.xml
+++ b/integration-tests/test-suite/partials/layout.xml
@@ -479,6 +479,8 @@
 		<collect>
 			<open/>
 			<resolution width="767"/>
+			<!--ToDo: Remove resolution-sleep-resolution workaround after issue #357 is fixed.
+			(https://github.com/Cognifide/aet/issues/357)-->
 			<sleep duration="2000"/>
 			<resolution width="767"/>
 			<screen css=".dynamic2"/>
@@ -495,6 +497,8 @@
 		<collect>
 			<open/>
 			<resolution width="767"/>
+			<!--ToDo: Remove resolution-sleep-resolution workaround after issue #357 is fixed.
+			(https://github.com/Cognifide/aet/issues/357)-->
 			<sleep duration="2000"/>
 			<resolution width="767"/>
 			<screen exclude-elements=".dynamic1,.dynamic2"/>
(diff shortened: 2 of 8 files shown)


core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/resolution/ResolutionModifier.java
```diff
@@ -41,7 +41,7 @@
 
   private static final String JAVASCRIPT_GET_BODY_HEIGHT = "return document.body.scrollHeight";
 
-  private static final int MAX_SIZE = 15000;
+  private static final int MAX_SIZE = 35000;
```
Skejven: Please update [Resolution Modifier wiki](https://github.com/Cognifide/aet/blob/master/documentation/src/main/wiki/ResolutionModifier.md#parameters).

integration-tests/test-suite/partials/layout.xml
```diff
@@ -474,5 +474,38 @@
...
+		</compare>
+		<urls>
+			<url href="comparators/layout/long_page.jsp"/>
+		</urls>
+	</test>
+
+	<test name="S-comparator-Layout-long-page-without-dynamic-at-bottom">
+		<collect>
+			<open/>
+			<resolution width="767"/>
```
Skejven: Could you please leave here `ToDo` comment with info that resolution-sleep-resolution workaround should be removed after #357 is fixed?

integration-tests/test-suite/partials/layout.xml
```diff
@@ -474,5 +474,38 @@
       <url href="comparators/layout/failed.jsp"/>
     </urls>
   </test>
+
+	<test name="F-comparator-Layout-dynamic-element-at-bottom-of-long-page">
```
Skejven: Could you please explain what is the idea for this failing test?

tkaik: I think it's about checking if AET are capable of taking partial screenshot of element located at the bottom (.dynamic2) of a very long page.
@Jakub-Izbicki I think you could also add a test (`F-` type) which takes the screenshot of the whole page without excluding the dynamic content at the bottom - this way we'll check if AET is capable of taking such big screenshots.

Jakub-Izbicki: @tkaik You are right about what the test does. Concerning your suggestion about additional test: I don't think it is necessary as, as far I noticed, AET is always doing a full page screenshot, regardless whether partial is present or not? Take a look:
```
  private byte[] takeScreenshot() throws ProcessingException {
    try {
      if (isSelectorPresent()) {
        SeleniumWaitHelper
            .waitForElementToBePresent(webDriver, getLocator(), getTimeoutInSeconds());
        return getImagePart(getFullPageScreenshot(), webDriver.findElement(getLocator()));
      } else {
        return getFullPageScreenshot();
      }
      . . .
```

integration-tests/test-suite/partials/layout.xml
Jakub-Izbicki: Done

integration-tests/test-suite/partials/layout.xml
tkaik: @Jakub-Izbicki You're right, I never noticed that :) However the full screenshot is taken but it's immediately changed to a partial-image - the full screenshot is not saved in the database or displayed on the report - I think it's good idea to test such case as well. What do you think?

Jakub-Izbicki: Good idea, fixed!

PR #397, opened 2018-10-16
Feature/inconsistent screenshot height

### LINKED ISSUE #357: Inconsistent screenshot height when height param in resolution modifier is omitted

Jakub-Izbicki:
Add algorithm to enable taking long screenshots without resolution-sleep-resolution workaround.
NOTE: This branch temporarily comes out of hotfix/screen-max-height as it uses 

## Description
This is a proposed solution no. 2 from issue #357. An algorithm that takes samples of page height in defined periods of time.
Adds new param to `<screen/>` called `samplingPeriod`, which defined waiting time between taking samples. 
Please give suggestions if all proposed default values in algorithm are ok, and if some of them should be configurable or not:
1. `samplingPeriod` - 100ms (configurable)
2. Number of last n height samples to compare - 3 (not configurable)
3. Max number of algorithm iterations = 15 (not configurable)

Please note, this pr is supposed to be merged after #387. There are several todos in code that indicate the tests should be changed to test proper page heights.

## Motivation and Context
Fixed #357 

## Types of changes

## Checklist:

---


core/jobs/src/main/java/com/cognifide/aet/job/common/utils/Sampler.java
```diff
@@ -0,0 +1,79 @@
...
+
+import com.cognifide.aet.job.common.modifiers.resolution.ResolutionModifier;
+import java.util.LinkedList;
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Sampler {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ResolutionModifier.class);
```
tkaik: Did you intentionally pass `ResolutionModifier.class` instead of `Sampler.class` to `getLogger` method here?

Jakub-Izbicki: Ups, copy-paste fail. Will fix that.

core/jobs/src/main/java/com/cognifide/aet/job/common/utils/Sampler.java
```diff
@@ -0,0 +1,78 @@
...
+   *
+   * @param samplesSupplier supplier of value to wait for,
+   * @param samplingPeriod milliseconds period between taking each sample,
+   * @param sampleQueueSize defines the last n elements that are to be compared,
+   * @param maxSamplesThreshold max number of samples before return
+   * @return true if pattern is empty or if its match with given value, false otherwise
+   */
+  public static <T> T waitForValue(Supplier<T> samplesSupplier, int samplingPeriod,
+      int sampleQueueSize, int maxSamplesThreshold) {
+    LinkedList<T> samplesQueue = new LinkedList<>();
```
tkaik: Did you consider using `org.apache.commons.collections.buffer.CircularFifoBuffer` ? I think you can use it here to simplify your code :)

eosiak:
Tested, OK!

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/resolution/ResolutionModifier.java
```diff
@@ -72,26 +85,37 @@ public void setParameters(Map<String, String> params) throws ParametersException
       if (params.containsKey(HEIGHT_PARAM)) {
         height = NumberUtils.toInt(params.get(HEIGHT_PARAM));
         ParametersValidator
-            .checkRange(height, 1, MAX_SIZE, "Height should be greater than 0 and smaller than " + MAX_SIZE);
+            .checkRange(height, 1, MAX_SIZE,
+                "Height should be greater than 0 and smaller than " + MAX_SIZE);
+      } else {
+        samplingPeriod = Optional.ofNullable(params.get(SAMPLING_PERIOD_PARAM))
+            .map(NumberUtils::toInt)
```
Skejven: `NumberUtils.toInt(str)` returns `0` when conversion fails.
Consider using `NumberUtils.toInt(str, default)`
```suggestion
            .map(spp -> NumberUtils.toInt(spp, DEFAULT_SAMPLING_WAIT_PERIOD);
```

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/resolution/ResolutionModifier.java
```diff
@@ -72,26 +85,37 @@ public void setParameters(Map<String, String> params) throws ParametersException
       if (params.containsKey(HEIGHT_PARAM)) {
         height = NumberUtils.toInt(params.get(HEIGHT_PARAM));
         ParametersValidator
-            .checkRange(height, 1, MAX_SIZE, "Height should be greater than 0 and smaller than " + MAX_SIZE);
+            .checkRange(height, 1, MAX_SIZE,
+                "Height should be greater than 0 and smaller than " + MAX_SIZE);
+      } else {
+        samplingPeriod = Optional.ofNullable(params.get(SAMPLING_PERIOD_PARAM))
```
Skejven: There is no limit for the `samplingPeriod`.
Just a question here: shouldn't that be limited somehow?
What will happen if user sets here e.g. `10000000 ms`?

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/resolution/ResolutionModifier.java
Jakub-Izbicki: Good catch, thank you.

core/jobs/src/main/java/com/cognifide/aet/job/common/modifiers/resolution/ResolutionModifier.java
Jakub-Izbicki: You are correct, there should be a range for this param. I will set max at 10 seconds. Please give suggestion if the value seems reasonable.

PR #422, opened 2018-10-31
RunnerConfiguration's urlPackageSize property got set to 1

### LINKED ISSUE #421: Fix default configuration of RunnerConfiguration
### Version used

* `AET 3.1.0`

### Context of the issue/feature

Default configuration [`com.cognifide.aet.runner.RunnerConfiguration.cfg`](https://github.com/Cognifide/aet/blob/3.1.0/osgi-dependencies/configs/src/main/resources/com.cognifide.aet.runner.RunnerConfiguration.cfg) contains invalid numbers.

- `urlPackageSize` which defines how many URL should be packed into a single collector task should be now set to `1`, as AET don't need to do the optimization and put more than 1 url in this packed at once any more due to change to Selenium Grid.

All fixes should be also applied to the default values in [RunnerConfigurationConf](https://github.com/Cognifide/aet/blob/master/core/runner/src/main/java/com/cognifide/aet/runner/configuration/RunnerConfigurationConf.java).

jwadolowski:
## Description

Fixes #421

## Motivation and Context

After comparison of AET's `config.zip` with settings that sit containerized version of it maintained by @Skejven it turned out there are differences that should be fixed upstream (details: Skejven/aet-docker/pull/5).

## Screenshots (if appropriate)

N/A

## Types of changes

## Checklist:

---


core/runner/src/test/java/com/cognifide/aet/runner/processing/steps/StepManagerTest.java
```diff
@@ -96,7 +96,7 @@ public void setUp() throws Exception {
     when(suiteIndexWrapper.get()).thenReturn(suite);
     when(suite.getCorrelationId()).thenReturn(CORRELATION_ID);
     when(runnerConfiguration.getMttl()).thenReturn(100L);
-    when(runnerConfiguration.getUrlPackageSize()).thenReturn(2);
+    when(runnerConfiguration.getUrlPackageSize()).thenReturn(1);
```
Skejven: This value shouldn't change. This unit test (actually this is abstract class so the test that inherits this class) checks if it is possible to set an `urlPackageSize`. And value `2` is perfectly fine here. As you will notice this will probably also fail unit tests and Travis build.

jwadolowski: Fixed

Skejven:
@jwadolowski could you please update changelog with this bugfix?

Issue #332, opened 2018-08-22
Support for CI/CD pipelines


mchrominski:
### Version used

* `AET 2.1.7`

### Context of the issue/feature

With the introduction of Shared Patterns (https://github.com/Cognifide/aet/issues/121) and later fix for using latest available 'patternSuite' we now have the higher comfort of AET usage within the development lifecycle.

However, it is still unnatural to use AETs within the CI pipeline. When a change is introduced and accepted on a single environment, if there are more environments in the pipeline, same change needs to be accepted on all the subsequent environments.

From my point of view, the test results (similar to expectations in the unit test methods) are closely linked to the source code and can be treated as a code signature. Surely, there are other factors like the environment configuration or the content that influence the tests, but I think it's reasonable to introduce *a* relation between a code (in given version) and the expected test results.

Similar use case appear in a hotfix process for an already release version. Currently, if a change is applied to an old release, the user first need to (blindly) accept the results of release retesting, then apply the change and look for any regression. Ideally, the first test outcome could be reused from the history making sure the change assessment is accurate.

After conducting couple of design sessions below are findings:
* The test results shouldn't be stored inside the repository, due to the large size
* Keeping the GIT graph (either code or commits) inside AET seems an abuse, due to tight coupling to this idea
* The feature should be an option, in some development lifecycles this isn't a real problem
* Quite often, there isn't a single GIT repository that is being deployed to the web server, due to the multi-module architectures
* Current Shared Patterns implementation is not enough, as the same pattern will be used for all the environments, but the expectation is that the environment (AET started for the environment) will detect the required pattern in a smart way
* The solution shouldn't base on commit hashes but rather on the state of files in the repository, due to merges being a separate commit
* If a task A has been accepted on the first environment, then task B, and the workflow allows task B to be merged before task A, it is expected that there will be no matching suite while testing the second environment, because such configuration was not tested before. Once task A is merged, the suite should be found
* There will be issues with regards to comparing results between environments, because the test outcome includes environment-specific content (like domain in URLs). This will be treated separately, with a goal to have an environment-free patterns
* Current implementation of pattern cleaner service is going to punish this solution so we'd need to have a smarter way of finding unused patterns, and the same time a way to mark a pattern as something to not to clean at all (because it's related to a LTS release)
* If a change is also to be introduced to the AET suite configuration, the repository should be part of the codebase meaning a change in suite also changes the state of repository binary state

### Steps to reproduce

1. Have two environments, being built from same repository but different branches
[...]

PR #424, opened 2018-10-31
Feature/sum control imp

WaldemarrGr:
Regarding to https://github.com/Cognifide/aet/issues/332. 
and for client side:  https://bitbucket.cognifide.com/projects/RND/repos/aet-checksum-generator/browse 

## Types of changes

## Checklist:

---


core/runner/src/test/java/com/cognifide/aet/runner/processing/data/SuiteDataServiceTest.java
```diff
@@ -0,0 +1,121 @@
...
+        .createSuite();
+    lastVersionSuite.setVersion(generateRandomPositiveLong());
+
+    suiteHashcode = new SuiteBuilder().setCorrelationId(patternCorrelationId)
+        .setCompany(COMPANY).setProject(PROJECT)
+        .setName(SUITE_NAME)
+        .setPatternCorrelationId("id")
+        .setProjectHashCode(CHECKSUM_01)
+        .createSuite();
+    service = new SuiteDataService(metadataDAO);//todo OSGI mock
```
WaldemarrGr: I have a problem with OSGI mock. At the moment mock is inject by constructor

client/client-scripts/aet.sh
```diff
@@ -44,6 +44,11 @@ function usage {
     exit 1
 }
 
+# todo pass login and password by parameters
+function get_json {
+	curl -u admin:admin -s "$1"
```
WaldemarrGr: This is a place for global variables (jenkins / bamboo) login/ password

test-executor/src/main/java/com/cognifide/aet/executor/SuiteServlet.java
```diff
@@ -104,6 +113,20 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
...
 
+  private String getChecksumFromJson(Map<String, String> requestData) {
+    String checkSum = StringUtils.EMPTY;
+    JsonParser parser = new JsonParser();
+    String projectChecksSum = requestData.get(PROJECT_CHECK_SUM);
+    JsonElement parse = parser.parse(projectChecksSum);
+    if (parse.isJsonObject()) {
+      JsonObject asJsonObject = parse.getAsJsonObject();
+      if (asJsonObject.isJsonObject()) {
+        checkSum = asJsonObject.get(PROJECT_CHECK_SUM).getAsString();
```
Jakub-Izbicki: The user will get NullPointerException here.
You are trying to extract checksum from `asJsonObject`, providing the key `"projectChecksSum"`, but the checksum property in the json which is returned by checksum servlet (/bin/checksum/combined)
you created in https://bitbucket.cognifide.com/projects/RND/repos/aet-checksum-generator/browse
is named `projectCheckSum` (check vs check**s**). So the `.get()` returns null and following `.getAsString()` throws NPE.

Due to this, the code in present form can never actually work.
The name of the property needs to be unified. Changed either in here or in `AetChecksumGenerator` project.

core/runner/src/main/java/com/cognifide/aet/runner/processing/data/SuiteDataService.java
```diff
@@ -15,38 +15,91 @@
...
+        pattern = metadataDAO.getSuite(dbKey, currentRun.getPatternCorrelationId());
+      }
     }
     return SuiteMergeStrategy.merge(currentRun, lastVersion, pattern);
   }
 
+  private Suite assignCheckSumToPattern(SimpleDBKey dbKey, Suite currentRun, Suite lastVersion, String checkSumCurrentRunProject) {
+    Suite pattern = null;
+    try {
+      if (isSuitHasPattern(currentRun)) {
```
Jakub-Izbicki: As far as I understand the code, if a current suite is provided with a `patternCorrelationId`, and with a current run `checksum`, this `checksum` will be assigned to the suite the`patternCorrelationId` is pointing to. 

Is this an expected behaviour? Due to this fact, multiple different codebases that are running tests with both these parameters passed will keep changing the `checksum` of the same suite in db. 

A situation could occur, when if a test would be run with only checksum parameter provided, the patterns may not be found, because some other environments could have changed the checksum, as they may have passed a `patternCorrelationId` of these patterns.

core/runner/src/test/java/com/cognifide/aet/runner/processing/data/SuiteDataServiceTest.java
Jakub-Izbicki: You can delete this line, annotate class with 
```
@RunWith(MockitoJUnitRunner.class)
```
and annotate service object field with
```
@InjectMocks
private SuiteDataService service;
```
and it will work.

Issue #419, opened 2018-10-30
Update documentation regarding Accessibility Collector


RdkLwg:
### Version used

* `AET 3.0`

### Context of the issue/feature

The page https://github.com/Cognifide/aet/wiki/AccessibilityCollector related to Accessibility does not seem to be valid. We should add a proper collector version and some more information regarding it.

PR #430, opened 2018-11-14
Upgrade codesniffer to 3.2.0

wblachowski:
I upgraded AccessSniff to version 3.2.0 (issue #420)

## Description
Htmlcs js has been updated to AccessSniff 3.2.0. All custom code has been incorporated into new version (the custom changes are listed in comment at the top of the file).

---
I hereby agree to the terms of the AET Contributor License Agreement.

PR #437, opened 2018-11-21
Feature/update accessibility collector doc

wblachowski:
Accessibility Collector wiki update (issue #419).

## Description
I added version of the AccessSniff used and removed "beta" information since the collector now uses stable sniffer.

---


Skejven:
For the future - it is best to update docs together with changes in the code, so #430 is a good place to contribute wiki pages update.

PR #442, opened 2018-11-27
Milestone/multiple patterns in test runs - wip

Jakub-Izbicki:
This is work in progress so everything may be subject to change, but I am uploading this after suggestion from @Skejven, for AET contributors to review and give feedback/suggestions.

Relates to #332 #424 

***General idea:***

The general idea behind these changes, is to allow user to provide multiple patternSuites/patternCorrelationIds when running suite, so the tests would be compared against not one but many patterns. This feature is actually based on already existing feature `Shared Patterns` but it expands it, allowing to provide more than one pattern references. 

Code from #424 will be eventually added to this functionality as well, which will basically provide yet another way in which user could pass references to shared patterns (by project checksum).

***Roadmap:***

Branch so far contains changes:
 - allow users to pass multiple patternSuite and patternCorrelationId params,
 - change java code structures to process collection of patterns through whole flow of running tests

To do:
 - introduce new structures in code to hold information about individual patterns (for example, so user can know what suite each individual pattern comes from, no longer a 1:1 relation),
 - adapt database to changes introduced in code,
 - implement frontend logic,
 - change frontend raport layout in a way it is able to present results with multiple patterns,
 - integrate with code from #424 

---

### DIFF
--- a/api/communication-api/src/main/java/com/cognifide/aet/communication/api/metadata/Comparator.java
+++ b/api/communication-api/src/main/java/com/cognifide/aet/communication/api/metadata/Comparator.java
@@ -29,7 +29,7 @@ public class Comparator extends Operation implements Commentable, Named {
 
   public static final String COMPARATOR_PARAMETER = "comparator";
 
-  private ComparatorStepResult stepResult;
+  private List<ComparatorStepResult> stepResults;
 
   private List<Operation> filters = new ArrayList<>();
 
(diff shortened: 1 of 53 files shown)


Jakub-Izbicki:
So far, there are changes in a lot of files, but most of the are quite straightforward, mostly converting structures to handle collection of patterns, instead of a single pattern. 

Some files that I would say require most scrutiny are:
`GetMetadataArtifactsProcessor.java`
`SuiteDataService.java`
`SuiteMergeStrategy.java`

PR #448, opened 2018-12-07
Feature/layout comparison multiple patterns

Jakub-Izbicki:
This is an attempt at implementing multiple shared patterns logic in the layout comparison tab.

Please review and give suggestions concerning code itself, but also give comments on how you think the layout comparison screen should look in the final version.

Here is a gif showing how it looks and works now, it is a result of a test run with provided two `-c` parameters (correlation ids of suites with patterns). The page is the one with dynamic content, so the test always fails. As seen in the image, comparisons with both two patterns failed, and patterns can each be viewed separately.

layout-comparison


report/src/main/webapp/app/services/metadata.service.js
```diff
@@ -149,6 +152,85 @@ define(['angularAMD', 'metadataCacheService', 'metadataEndpointService'],
...
+            }
+          });
+
+          return isAnyAcceptable;
+        }
+
+        function isRebaseable() {
+          var isAnyRebaseable = false;
+
+          _.forEach(stepResults, function (stepResult) {
```
Skejven: You don't need to iterate to the end of the loop. You need to find first `stepResult` that is rebaseable.
Try to use `find` insead of `forEach`:

```js
var rebaseableStepResult = _.find(stepResults, function (stepResult) {
  return stepResult.rebaseable;
});
return !!rebaseableStepResult;
```

report/src/main/webapp/app/services/metadata.service.js
```diff
@@ -149,6 +152,85 @@ define(['angularAMD', 'metadataCacheService', 'metadataEndpointService'],
...
+        function setStatus(newStatus) {
+          _.forEach(stepResults, function (stepResult) {
+            stepResult.status = newStatus;
+          });
+        }
+
+        function isAcceptable() {
+          var isAnyAcceptable = false;
+
+          _.forEach(stepResults, function (stepResult) {
```
Skejven: See my comment for `isRebaseable()`.

report/src/main/webapp/app/services/metadata.service.js
```diff
@@ -149,6 +152,85 @@ define(['angularAMD', 'metadataCacheService', 'metadataEndpointService'],
...
+            }
+          });
+
+          return foundStatus;
+        }
+
+        function anyResultHasStatus(status) {
+          var hasStatus = false;
+
+          _.forEach(stepResults, function (stepResult) {
```
Skejven: See my comment for `isRebaseable()`.

report/src/main/webapp/app/services/metadata.service.js
Jakub-Izbicki: Thanks! I fixed the code, also in other places you mentioned too.

report/src/main/webapp/assets/sass/_report.scss
```diff
@@ -226,6 +226,91 @@
...
+.toolbar-patterns-base {
+  border-bottom: 2px solid #dcdce5;
+  padding: 10px;
+  margin: 0 -25px 10px -25px;
+  width: 110%;
+}
+
+.toolbar-patterns-header {
+  font-family: montserratlight, sans-serif;
+  color: #77777c;
```
plutasnyy: Maybe you could use here some predefined variable? 😃

report/src/main/webapp/app/services/metadata.service.js
```diff
@@ -246,12 +338,19 @@ define(['angularAMD', 'metadataCacheService', 'metadataEndpointService'],
...
           if (decoratedObject.comparators) {
             var revertedPatterns = 0;
             _.forEach(decoratedObject.comparators, function (comparator) {
               if (hasStepComparatorAcceptedChanges(comparator)) {
-                comparator.stepResult.status = comparator.stepResult.previousStatus;
-                comparator.stepResult.previousStatus = null;
+                _.forEach(comparator.stepResult.getStepResults(),
+                    function (stepResult) {
+                      console.log("Accepting comparator changes");
+                      console.log(comparator);
```
plutasnyy: Is it needed?

report/src/main/webapp/app/services/patterns.service.js
```diff
@@ -151,13 +160,15 @@ define(['angularAMD', 'metadataService', 'metadataAccessService'],
         }
 
         function isCollectorWithPattern(step) {
-          return step && step.comparators && step.pattern;
+          return step && step.comparators &&
+              step.patterns &&
+              step.patterns.length > 0;
```
plutasnyy: Is it correctly formatted?

report/src/main/webapp/assets/sass/_report.scss
Jakub-Izbicki: Done, thanks :P

report/src/main/webapp/app/services/metadata.service.js
Jakub-Izbicki: Good catch! I removed this logging.

report/src/main/webapp/app/services/patterns.service.js
Jakub-Izbicki: I reformatted this code with google formatter again, and now it looks different :)

Issue #436, opened 2018-11-21
Ambiguous message presented when URLs are duplicated or empty


artswie:
If in single suite will exist empty or duplicated URLs (this might happen fairly easy on huge suites) you will get ambiguous messages while running "malformed" xml. 

The message presented on jenkins on older version of aet (2.1.3):
`[ERROR] Failed to process test suite: 'Not Found'`

The message presented on jenkins on lastest version of aet (3.1.0):
`[ERROR] Failed to schedule test suite: '[Status: 400] Bad Request`

Please provide ERROR message which will be more informative OR prevent it in some other way. 

Please run below code and unncomenty one of urls to see a failure.
```
<?xml version="1.0" encoding="UTF-8"?>
<suite name="testing" company="cognifide" project="cog" domain="https://www.cognifide.com">
  <test name="Accessibility" useProxy="rest">
    <collect>
	  <resolution width="1920"/>
      <open/>  
	  <accessibility standard="WCAG2AAA" />  
    </collect>
    <compare>
      <accessibility report-level="ERROR" />
    </compare>
	<urls>		
		<url href="/contact-us"/>
		<!--<url href=""/>-->
		<!--<url href="/contact-us"/>-->
	</urls>
	</test>
</suite>
```

Cheers!

radoslawkarwacki:
I looked into it and made a little improvement (will create pull request shortly). 

The logs for this set of URLs:

    <urls>      
        <url href=""/>
        <url href="/test0"/>
        <url href="/test0"/>
        <url href="/test1" name="name-1"/>
        <url href="/test1" name="name-1"/>
        <url href="/test2" name="name-2"/>
        <url href="/test2" name="name-3"/>
        <url href="/test3" description="description-1"/>
        <url href="/test3" description="description-1"/>
        <url href="/test4" description="description-2"/>
        <url href="/test4" description="description-3"/>
    </urls>

Will be:

```
[Status: 400] Something is wrong with your suite definition! Detected errors: 
Empty url
Duplicated url, with URL: /test0, with name: /test0
Duplicated url, with URL: /test1, with name: name-1
Duplicated url, with URL: /test3, with name: /test3, with description: description-1
Duplicated url, with URL: /test4, with name: /test4, with description: description-3

Ideally, this should also contain line numbers. I didn't dig deep into it, but I'm not sure if the Simple XML Framework library can provide them. That should be considered as another improvement.

[...]

PR #459, opened 2018-12-21
Print more meaningful error messages when suite.xml is malformed

radoslawkarwacki:
Print more meaningful error messages when suite.xml is malformed

## Description
Provide as much data as possible when URLs are empty or duplicated

## Motivation and Context
https://github.com/Cognifide/aet/issues/436

## Types of changes

## Checklist:

---


plutasnyy:
Please update changelog :)

bambroziak:
Tested and rejected.

There is an issue there:
While I specify two tests on AET suite.xml and in both of them I've malformed (duplicated, left empty urls), I've got output only by first test, there is no any info about second test despite that it is also broken.

Other scenario is while I fix first test and leave second one broken, I've got output without information about that errors are located in test number 2. If first test is broken there is displaying info like that.

Expect of issues above error messages are working fine and displays useful data.

QA Covered.

tkaik:
@bambroziak Issue you desribed is acceptable - this improvements is made on existing mechanism which reports error only about the first faulty <test> of given suite. Reporting errors of all tests or reporting the faulty test name would require changing this mechanism which doesn't need to be done in scope of this PR.

PR #460, opened 2018-12-24
Unescaping already escaped xml URLs

### LINKED ISSUE #441: XML attribute values in suite.xml are not un-escaped
**Describe the bug**
When the suite XML file contains attributes with escaped special characters (e.g. `&` escaped to `&amp;` in url's `href` attribute), AET does not un-escape such value. Therefore it forces us to write unescaped special characters in suite file to achieve expected behaviour, which makes them invalid XML.

* Version used: `AET 3.1.0`

**To Reproduce**
1. Prepare a suite.xml file with url's href attribute with multiple query string parameters (to use `&` sign), e.g. `<url href="https://en.wikipedia.org/wiki/Main_Page?a=b&amp;c=d"/>` (you may modify example suite from https://github.com/Cognifide/aet/wiki/AETIn10Minutes#set-up-the-test )
2. Run the suite (e.g. `./aet.sh http://localhost:8181 suite.xml`)
3. Open the report and navigate to the example URL with query parameters
4. See that the URL is displayed with `&amp;` - it should have been un-escaped to `&` by AET. Also if you click on the URL displayed in the report, it takes you to the page with `&amp;` in query string.

**Expected behavior**
AET should un-escape all escaped XML characters in the suite xml. I would also recommend that AET should only accept suite XML files which are valid XML files.

**Screenshots**
2018-11-27_11h11_18

**Additional context**

wblachowski:
I made a change fixing unescaped URLs in report page. For details see #441.

## Description
Current implementation expects unescaped URLs in suite XML (e.g `(...)?a=b&c=d` ) which are then escaped before serializing XML to TestSuite. If you put correctly escaped XML in URL (e.g `(...)?a=b&amp;c=d`) then the query params are escaped to `(...)?a=b&amp;amp;c=d`, parsed back (while serializing) to `(...)a=b&amp;c=d` and this version is then displayed in report page.

This redundancy in escaping is what causes the issue. If the URLs are already escaped they shouldn't be tampered with. My fix uses the simpliest solution, i.e it tries to unescape before escaping, thus handling correctly both unescaped (unescape has no effect here) and escaped URLs. 

I also added a unit test for escaped URLs.
## Motivation and Context
Closes #441 

## Types of changes

## Checklist:

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ All notable changes to AET will be documented in this file.
 - [PR-404](https://github.com/Cognifide/aet/pull/404) Added missing tooltip for conditional tests
 - [PR-408](https://github.com/Cognifide/aet/pull/408) Advanced Screen Comparision button layout fix
 - [PR-410](https://github.com/Cognifide/aet/pull/410) Notification that displays when exclude-elements are not found on page now shows what specific elements were not found([#372](https://github.com/Cognifide/aet/issues/372)) 
+- [PR-460](https://github.com/Cognifide/aet/pull/460) Unescaping already escaped xml URLs ([#441](https://github.com/Cognifide/aet/issues/441)) 
 
 ## Version 3.1.0
 
--- a/test-executor/src/main/java/com/cognifide/aet/executor/xmlparser/xml/utils/EscapeUtils.java
+++ b/test-executor/src/main/java/com/cognifide/aet/executor/xmlparser/xml/utils/EscapeUtils.java
@@ -49,7 +49,7 @@ public final class EscapeUtils {
 
         if (attr.startsWith("href=")) {
           attr = String.format("href=\"%s\"",
-              StringEscapeUtils.escapeXml11(attr.substring(6, attr.length() - 1)));
+              StringEscapeUtils.escapeXml11(StringEscapeUtils.unescapeXml(attr.substring(6, attr.length() - 1))));
         }
         attrMatcher.appendReplacement(attrsStringBuffer, attr + " ");
       }
--- a/test-executor/src/test/java/com/cognifide/aet/executor/xmlparser/xml/utils/EscapeUtilsTest.java
+++ b/test-executor/src/test/java/com/cognifide/aet/executor/xmlparser/xml/utils/EscapeUtilsTest.java
@@ -28,6 +28,14 @@ public class EscapeUtilsTest {
 
   @Test
   public void escapeUrls_expectEscapedResults() throws IOException {
+    String xmlString = readContentFromFile("/testSuite_unescapedUrls.xml");
+    String escapedXmlString = EscapeUtils.escapeUrls(xmlString);
+    String resultString = readContentFromFile("/testSuite_escapedUrls_result.xml");
+    assertThat(escapedXmlString, is(resultString));
+  }
+
+  @Test
+  public void escapeUrls_alreadyEscaped_expectUnchangedResults() throws IOException{
     String xmlString = readContentFromFile("/testSuite_escapedUrls.xml");
     String escapedXmlString = EscapeUtils.escapeUrls(xmlString);
     String resultString = readContentFromFile("/testSuite_escapedUrls_result.xml");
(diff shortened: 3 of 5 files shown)


wblachowski:
Hello @wiiitek, you've made a few good points and I feel obliged to clear up some misunderstandings.
First of all:
>But something is not right here. Isn't it that:
>StringEscapeUtils.escapeXml11(StringEscapeUtils.unescapeXml( something )) === something
>??

That's true only for escaped strings. Take for example this string: `&`. The above statement evaluates to `escape(unescape(&)) == escape(&) == &amp; =/= &`.
As you can see `unescape()` has no effect on not-escaped strings. This little workaroud of unescaping and escaping ensures that at the end we get an escaped string regardless of initial input being escaped or not.

Secondly:
>But what should we do with this URL:
>        `<url href="https://www.google.pl/search?param=12&q=&quot;" />`
>If something like that is placed inside suite - what page should AET test? Currently it will do search for >&quot;, but after the proposed change it will search for "?

You are right that this url will in the end be serialized to `(...)?param=12&q="`, however I have some problems with this url. To me, it would never be necessary to use XML escaping on query strings. If you have a valid URL, all XML special characters (`&<>"'`) are already escaped. The exception is `&` delimiting query parameters

In conclusion:
To my mind the only enemy among XML special characters (`&<>"'`) is the ampersand because I don't see any possibility of any of those `<>"'` occurring unescaped in a valid URL, although I might be wrong. With a valid URL user has to decide whether to escape all `&` chars with `&amp;` or keep them unescaped (thus making suite.xml invalid XML file). Both of those options will be correctly serialized but I agree that it might be misleading and sticking to one option would be nice.

wiiitek:
Thank you, yes I agree :

> To me, it would never be necessary to use XML escaping on query strings. If you have a valid URL, all XML special characters (&<>"') are already escaped. The exception is & delimiting query parameters

But it might be confusing what one should do in case of testing escaped special characters on search pages. Maybe we should update documentation? ;p
What should i put into XML if i want to test something like that:

escaped query string

?

wblachowski:
**Dealing with `<>"`**
Hmm, I delved deeper into the subject and according to [RFC1738](https://www.ietf.org/rfc/rfc1738.txt) characters such as  `<>"` are regarded unsafe and should ALWAYS be encoded.

In fact, in your screenshot they are escaped but it appears that chrome tries to be user friendly and displays prettified version in the url bar. Fortunately, if you copy the address in the url bar and paste it in a text editor you get escaped version.
Example: 
the link you provided: 
https://www.google.pl/search?source=hp&q="%26amp%3B+vs+%2526"
copied from the url bar into notepad looks like this:
https://www.google.pl/search?source=hp&q=%22%26amp%3B+vs+%2526%22

I carried out a small experiment and pasted 
http://nonexistingpage.mock/?checkmark=%E2%9C%93&q=%22heh%22 
in url bars of different browsers. Results:

browsers
If you copy it back from the url bar to the notepad you get the original (escaped) URL.
Concluding, i think that unless you type the url by hand you shouldn't have to deal with escaping query parameters.

**Dealing with `;`**

The only problem left is `;` character. Similarly to `&` it is a reserved character and thus can occur unescaped in URLs. We wouldn't want strings such as `&amp;`, `&lt;` existing in valid URLs because they would get wrongly interpreted. 
W3 reccommends ([here](https://www.w3.org/TR/1999/REC-html401-19991224/appendix/notes.html#h-B.2.2)) the possiblity of using `;` as a query string separator instead of `&`.
So i guess a set of query strings like this: "a","amp","b" can IN THEORY be represented as `?a&amp&b` or `?a;amp;c` or even `?a&amp;c` (this is a a strange mix-up but i guess it's valid?). This remains a problem but it seems like an edge case almost impossible to occur.

I have to admit this is more complicated than it seemed :p I wonder what are your thoughts

wiiitek:
Thanks @wblachowski for your answer! I see you put an effort to look into this.

Please make a decision on this topic after some consultations with @Skejven and @mchrominski.

Personally I like the approach with with "valid XML" and making no escaping/unescaping but i understand that this could be not acceptable because of end-users. My other thought is that the XML we are using is not friendly format - we don't have XML schema (I remember there were attempts to prepare one). So any other format would be better ;p

Thank you again for your work. I believe you will make good decision. Cheers!

dzuma:
Whole this topic is resolved here: https://github.com/Cognifide/aet/pull/524
I think this PR can be closed.

PR #461, opened 2018-12-24
Generating correct correlationId when suite name is overridden

### LINKED ISSUE #440: The correlationID is not generated correctly when suite name is overriden
**Describe the bug**
Suite correlationID is incorrectly generated when the **suite name** is overriden by a client script parameter (e.g. `--name` parameter when using `aet.sh`). The generated correlationId always contains the original suite name from the XML file, instead of the overriden suite name.

* Version used: `AET 3.1.0`

**To Reproduce**
1. Create a new, valid suite.xml file (e.g. use the example from https://github.com/Cognifide/aet/wiki/AETIn10Minutes#set-up-the-test)
2. Run it with client script parameter whichs overrides the suite name, e.g. `./aet.sh http://localhost:8181 --name replaced-suite-name`
3. Open the generated report - see that the correlationId contains the original suite name from XML file.

**Expected behavior**
The correlationId should be generated using the overriden suite name. In general, when the suite name is provided as the client script parameter, the whole AET suite processing should ignore the original suite name from the XML file.

**Screenshots**

**Additional context**

wblachowski:
I fixed generating correlationId with suite name overridden in command line. For details see #440.

## Description
CorrelationId now contains overridden suite name instead of the value from XML file.

## Motivation and Context
Closes #440

## Types of changes

## Checklist:

---


test-executor/src/main/java/com/cognifide/aet/executor/model/TestSuiteRun.java
```diff
@@ -71,7 +71,7 @@ public TestSuiteRun(TestSuiteRun testSuiteRun, String name, String domain, List<
     this.company = testSuiteRun.getCompany();
     this.project = testSuiteRun.getProject();
     this.domain = domain;
-    this.correlationId = testSuiteRun.getCorrelationId();
+    this.correlationId = CorrelationIdGenerator.generateCorrelationId(company, project, name);
```
tkaik: It looks like this class has 2 constructors and now the 2nd constructor could use the 1st one to reduce the amount of code, what do you think? E.g. 2nd constructors could be changed to:
```
public TestSuiteRun(TestSuiteRun testSuiteRun, String name, String domain, List<TestRun> tests) {
    this(name, testSuiteRun.getCompany(), testSuiteRun.getProject(), domain, tests);
  }
```

wblachowski: Great suggestion, it seems like a reasonable thing to do.

tkaik:
Can you add some unit tests for TestSuiteRun or SuiteExecutor (see which fits better) which test if correlation ID is correct when the suite name is overriden? Just to make sure that noone will break this again accidentally when doing some refactoring :)

PR #488, opened 2019-02-28
Fix SuiteRemoveCondition unit test

wblachowski:
I fixed a SuiteRemoveConditionTest class which led to occasional tests failure.

## Description
I noticed that running SuiteRemoveConditionTest caused different results each time. Out of 100 runs, 87 ended successfully and 13 failed. Each fail was caused from assertion error in `evaluate_when8SuitesRemoveOlderThanButAlwaysKeepLatestVersion_expectTrue()` when testing the following suite list:  `A-1,A-2,B-3,B-4,C-5,C-6,D-7,E-8 ; 4 ; null ; C-5 ; 4`.
In this test we check if suite created 4 days ago is marked as "to be removed" if cleaner removes suites 4 days old and older.

The problem resulted from incorrect order of initalizations: first we create SuiteRemoveCondition with max acceptable timestamp (now minus 4 days) and then mocked Suite (with creation date: now minus 4 days). The offset between "now" value in these cases led to the Suite being considered younger than 4 days from the condition's perspective. The difference is so small that in some cases the test passes, but the bug is evident if you use a breakpoint between the two initializations (increasing the time difference - then it fails every time).

I changed the order in all methods to prevent future mistakes.
## Types of changes

## Checklist:

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ All notable changes to AET will be documented in this file.
 - [PR-413](https://github.com/Cognifide/aet/pull/413) Added ability to show full page source when there has been no difference discovered ([#369](https://github.com/Cognifide/aet/issues/369))
 
 - [PR-463](https://github.com/Cognifide/aet/pull/463) Removed dependency to Joda time library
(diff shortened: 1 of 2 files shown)


plutasnyy:
Update Changelog pls ;)

PR #489, opened 2019-03-07
Cleaner integration tests

wblachowski:
Added cleaner integration tests as suggested in #441

## Description
My solution uses a [mocked osgi context](https://sling.apache.org/documentation/development/osgi-mock.html) as well as [in-memory mock of mongoDB server](https://github.com/bwaldvogel/mongo-java-server). It checks various combinations of cleaner params (versions to keep and suite max age) and verifies whether correct metadata and artifacts documents have been removed from database.

For testing data I used real json objects extracted from my local aet instance db after a couple of runs (with some timestamp changes).

To make testing possible I had to incorporate some changes in cleaner code. This includes `LocalDateTimeProvider` interface to mock current system time and `CamelContextCreator` to inject different camel context configuration during testing and normal execution.

## Motivation and Context
Refers to enhancement brought up in #441

## Types of changes

## Checklist:

---


plutasnyy:
Good job!

core/cleaner/pom.xml
```diff
@@ -33,6 +33,22 @@
   <name>AET :: Core :: Cleaner</name>
   <description>Cleaner responsible for removing outdated AET artifacts from datastorage</description>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
```
tkaik: If those are dependencies only for tests then we should probably add`<scope>test</scope>`

wblachowski: done :wink:

integration-tests/cleaner-test/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+
+  <parent>
+    <artifactId>aet-root</artifactId>
+    <groupId>com.cognifide.aet</groupId>
+    <version>3.2.3-SNAPSHOT</version>
+    <relativePath>../..</relativePath>
+  </parent>
+
+  <artifactId>cleaner-test</artifactId>
+  <packaging>jar</packaging>
```
Skejven: Wy do we need `jar` distribution of this module?
It is only for tests.

integration-tests/cleaner-test/src/main/java/com/cognifide/aet/cleaner/InMemoryDB.java
```diff
@@ -0,0 +1,99 @@
...
+import com.mongodb.client.MongoDatabase;
+import de.bwaldvogel.mongo.MongoServer;
+import de.bwaldvogel.mongo.backend.memory.MemoryBackend;
+import java.io.File;
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import org.apache.commons.io.FileUtils;
+import org.bson.Document;
+
+public class InMemoryDB {
```
Skejven: Shouldn't all files in this module be in `src/test`?
Why do we need any of the code to be non-test?

integration-tests/cleaner-test/pom.xml
tkaik: Yes, if we set it to `pom` then the tests are not being compiled or executed

integration-tests/cleaner-test/src/main/java/com/cognifide/aet/cleaner/InMemoryDB.java
tkaik: You're right, I moved all files to `src/test`

PR #504, opened 2019-04-10
Cleaner redesign [REQUIRES WORK]

wblachowski:
Incomplete cleaner logic redesign. The changes include restoring [old cleaner version](https://github.com/Cognifide/aet/blob/master/documentation/src/main/wiki/releases/2.0.0/assets/diagrams/aet-cleaner-workflow.png) from AET 2.0.0 and adding a new cleaner service designed specially to remove orphaned artifacts from database.

## Description
The proposed solution separates two different cleaning tasks:

- **Expired suites+artifacts cleaning** - removes suites that are too old (either by days or by version number). Any artifacts referenced only by these suites should be deleted as well.
- **Orphaned artifacts cleaning** - removes artifacts not referenced by any existing suite. The possibility of orphaned artifacts existing in a database is a result of the new rerun feature.

These two tasks require different approaches so they are handled by two different cleaners, each with a separate CRON expression:
- **MetadataCleaner** - it's the same as the old cleaner from AET 2.0.0. It marks each metadata document with a flag stating whether is should be removed.  Apart from that it collects all artifacts that should be kept and all that should be removed. Then it proceeds to remove unwanted metadata and artifacts (by calculating the difference between a set of artifacts to be removed and a set of artifacts to keep).
- **OrphanCleaner** - this cleaner deals with orphaned artifacts only. Removing orphans is a bit tricky because it requires gathering a set of all referenced artifacts and then deleting everything not present in this set. This approach "deletes by default" so it is very important not to skip any relevant artifact while fetching referenced artifacts. Because of the way AET processes its tests, artifacts are inserted into db before a metadata document referencing them is saved. This could lead to inadvertent removal of valid artifacts. Because of that, OrphanCleaner uses a **global lock on suite execution** instance to ensure that no artifacts are added during its work. 

This PR still requires some work in **OrphanCleaner**. Currently there is no mechanism to handle OrphanCleaner starting during execution of any test. Those tests are not affected by the block mentioned above and this can lead to unwanted removal of artifacts gathered by them.

## Motivation and Context
This change is intended to solve recurring problems with the cleaner module, the most important being the unwanted removal of some artifacts. By providing two different cleaners with two separate CRON expression it enables users to run non-blocking cleaner (MetadataCleaner) more frequently than blocking cleaner (OrphanCleaner).

## Types of changes

## Checklist:

---


tkaik:
Please update Cleaner documentation https://github.com/Cognifide/aet/blob/master/documentation/src/main/wiki/Cleaner.md

core/cleaner/src/main/java/com/cognifide/aet/cleaner/configuration/CleanerSchedulerConf.java
```diff
@@ -39,9 +43,20 @@
...
-  String schedule();
+      name = EXPIRED_SCHEDULE_CRON,
+      description = "CRON notation of when the Expired Data Cleaner job is to be fired. [example: '0 0 21 ? * *' will trigger job daily at 21:00].")
+  String expiredCleanerSchedule();
+
+  @AttributeDefinition(
+      name = ORPHAN_RUN,
+      description = "Flag that says if Orphan Cleaner should be run with a given schedule.",
+      type = AttributeType.BOOLEAN)
+  boolean runOrphanCleaner() default false;
```
tkaik: Please add some example config for orphan cleaner in cfg files under `osgi-dependencies/configs/src/main/resources/`

integration-tests/cleaner-test/pom.xml
```diff
@@ -42,10 +42,47 @@
         <version>1.12.0</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>org.awaitility</groupId>
+        <artifactId>awaitility</artifactId>
+        <version>3.1.6</version>
+        <scope>test</scope>
+      </dependency>
+      <dependency>
```
tkaik: Do you need to add mockito, junit and hamcrest here in dependencyManagement section? Those libraries are already configured in root pom.xml (root repository folder)

PR #507, opened 2019-04-29
Fuzzy pixel color comparison

kaczymuczy:
## Description
Introduce an option to match colors that are close to the target color in a grayscale space when comparing pixels with the LayoutComparator.
The comparison can now run in two modes:
- the standard mode (nothing changed),
- "fuzz mode" - the baseline pattern and new screenshot images are converted to grayscale and the comparison takes into account a "fuzz" - a value that describes how different (how big is the distance) the pattern and screenshot pixels can be in order to consider them equal.
In other words - a naive implementation of https://imagemagick.org/script/command-line-options.php#fuzz

## Motivation and Context
This option allows to filter out image rendering issues like AEM image rendition differences not visible to the human eye.

## Types of changes

## Checklist:

---


kaczymuczy:
Testing on a working AET instance not done yet.

kaczymuczy:
Documentation not done yet.

kaczymuczy:
Testing on a working AET instance - done, works fine.

kaczymuczy:
Documentation done

wiiitek:
We could add some slightly different images (lie those generated by AEM renditions engines) to our integration tests with example of how to use `fuzz` parameter in test suite.
What do you think about this idea?

kaczymuczy:
https://github.com/Cognifide/aet/pull/507#issuecomment-487946320
The unit tests pretty much cover everything there is to test for this feature and the documentation covers the usage example. I don't think extra integration tests would bring more value - you believe this is necessary?

wiiitek:
https://github.com/Cognifide/aet/pull/507#issuecomment-489588976
Thank you for your response!

I don't think they are really necessary. And also I don't know what is the role of integration tests currently.

I have looked into sample files for integration tests - they are taken from the real-life use cases, right?

plutasnyy:
I like this feature very much! Problems with images were really really annoying. I think a bit about users - if a test is marked as a fail user doesn't know if it was because of threshold or fuzzy. I also think that using it together is not the best way so I wouldn't change anything.
The second thing that I wouldn't know is how big is the difference between some values of fuzzy, maybe some comparison of noised images that pass a test with the fuzzy value equal to 5% 10% 15% etc could be helpful?
Please update changelog :D
Good job!

kaczymuczy:
https://github.com/Cognifide/aet/pull/507#issuecomment-490006571
Yes, the samples are built from a number of real life cases.

Changelog updated

So, what exactly am I required to do to have this merged? I would love to use this as soon as possible. :D

Issue #515, opened 2019-07-18
Header Modifier doesn't work for "User-Agent" property


mniezgodka:
Header Modifier doesn't work for "User-Agent" property.

To reproduce:
1) Execute below suite
[suite.zip](https://github.com/Cognifide/aet/files/3407023/suite.zip)

Expected:
"User-Agent" property is being override with value: "Mozilla/5.0 (Windows NT 6.1; WOW64; rv:33.0) Gecko/20120101 Firefox/33.0"
and browser is recognized as: "You’re using Firefox 33."

Actual
"User-Agent" property isn't override and browser is recognized as:  "You’re using Chrome 75."
![2019-07-18_16h35_07](https://user-images.githubusercontent.com/42374634/61466780-cce25500-a97a-11e9-953a-21a4550bb062.png)

PR #516, opened 2019-07-29
Bugfix/header modifier doesn't work for user agent property

Cich0sza:
## Description

The problem with changing some headers like User-agent, Accept was solved by changing proxy server configuration. 

## Motivation and Context

#515 

## Types of changes

## Checklist:

---

### DIFF
--- a/core/jobs/src/test/java/com/cognifide/aet/job/common/modifiers/header/HeaderModifierTest.java
+++ b/core/jobs/src/test/java/com/cognifide/aet/job/common/modifiers/header/HeaderModifierTest.java
@@ -58,7 +58,7 @@ public class HeaderModifierTest {
     headerModifier.setParameters(ImmutableMap.of("key", "header", "value", "value1"));
     headerModifier.collect();
 
-    verify(proxyServer).addHeader("header", "value1");
+    verify(proxyServer).addHeader("header", "value1", false);
     verify(requestExecutor).addHeader("header", "value1");
   }
 
--- a/documentation/src/main/wiki/HeaderModifier.md
+++ b/documentation/src/main/wiki/HeaderModifier.md
@@ -14,6 +14,7 @@ In order to use this modifier it must be declared before the open module in the 
 | --------- | ----- | ----------- | --------- |
 | `key` | x | Key for the header | yes |
 | `value` | y | Value for the header | yes |
+| `override` | `true`/`false` | Replace existing header instead of adding new one | no (default false) |
 
 ##### Example Usage
 
@@ -23,6 +24,9 @@ In order to use this modifier it must be declared before the open module in the 
     <test name="header-modify-test" useProxy="rest">
         <collect>
             ...
+            <header key="User-Agent" value="Mozilla/5.0 (Windows NT 6.1; WOW64; rv:33.0) Gecko/20120101 Firefox/33.0" 
+                    override="true"/>
+            <header key="Accept" value="application/json" override="true"/>
             <header key="Authorization" value="Basic emVuT2FyZXVuOnozbkdAckQZbiE=" />
             ...
             <open />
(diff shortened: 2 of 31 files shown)


osgi-dependencies/proxy/src/main/java/com/cognifide/aet/proxy/RestProxyServer.java
```diff
@@ -95,18 +94,10 @@ public void setCaptureContent(boolean captureContent) {
...
 
   @Override
-  public void addHeader(String name, String value) {
+  public void addHeader(String name, String value, Boolean override) {
     CloseableHttpClient httpClient = HttpClients.createSystem();
     try {
-      URIBuilder uriBuilder = new URIBuilder().setScheme(HTTP).setHost(server.getAPIHost())
-          .setPort(server.getAPIPort());
-      // Request BMP to add header
-      HttpPost request = new HttpPost(uriBuilder.setPath(
```
Skejven: According to https://github.com/lightbody/browsermob-proxy#rest-api docs:
`POST    /proxy/[port]/headers`
> Set and override HTTP Request headers

Do those changes mean, that this is not true? There is even an example of setting the `User-Agent` header.

documentation/src/main/wiki/HeaderModifier.md
```diff
@@ -14,6 +14,7 @@ In order to use this modifier it must be declared before the open module in the
 | --------- | ----- | ----------- | --------- |
 | `key` | x | Key for the header | yes |
 | `value` | y | Value for the header | yes |
+| `override` | `true`/`false` | Replace existing header instead of adding new one | no (default false) |
```
Skejven: What's the case of using the Header Modifier without `override=true`?
Is that the case, that the user wants to have more than one header's value?
If that's the case, `override=true` should be the default option and eventually let the user add more header values using `override=false`.
Could you please also make it more clear, in the docs, that `override=false` means, that if the header already occurs in the request, it will not be modified (like with `User-Agent` example, where Chrome instance sets it).

Skejven:
What also concerns me is that https://github.com/lightbody/browsermob-proxy is no longer maintained:
image
There are over 160 issues open at the moment. If the docs for headers is not up-to-date maybe we should consider changing BMP for some different proxy?
However, this is a topic for another discussion and PR.

osgi-dependencies/proxy/src/main/java/com/cognifide/aet/proxy/RestProxyServer.java
Cich0sza: Yes, this doesn't work anymore. You can't use this to change some headers like User-agent. Using this you can only add new header or override header that you add before.

documentation/src/main/wiki/HeaderModifier.md
Cich0sza: Override false is use when you want add new header, true option is used to change headers that are add by Chrome instance. This is why false is default strategy.

I'll make documentation more clear about that.

core/jobs/src/test/java/com/cognifide/aet/job/common/modifiers/header/HeaderModifierTest.java
```diff
@@ -58,7 +58,7 @@ public void testCollect() throws Exception {
     headerModifier.setParameters(ImmutableMap.of("key", "header", "value", "value1"));
     headerModifier.collect();
 
-    verify(proxyServer).addHeader("header", "value1");
+    verify(proxyServer).addHeader("header", "value1", false);
```
Skejven: Is there any way to test the `override` approach?
Maybe if not the unit test, some integration test (e.g. page displaying `User-Agent` header value?)

PR #517, opened 2019-08-01
URL rerun

### LINKED ISSUE #487: URL rerun doesn't work
**Describe the bug**
Rerun url doesn't work

* Version used: `AET 3.2.0`

**To Reproduce**
1. Go to any URL
2. Click on rerun url button

You will see:
```
Status: URL rerun initialized

Progress: 0.00%
```
But it never change, I also see in "network" that URL in request looks good, every necessary parameter is added. In addition Rerun Test works correctly and URL is the same (without information about url to rerun obviously). The `currentRerunUrlEndpoint` in localStorage is `undefined`, in my opinion there could be a problem, but I am a bit confused because request is correct but after refresh I don't see information about the last rerun so probably it wasn't sent to process? It is strange, but I didn't investigate it more preciously.

**Expected behavior**
URL rerun works

Cich0sza:
##  Description

Angular app was sending a url instead of a url name, so backend could not find the right test case.
- Angular app sends a url name instead of a url.

## Motivation and Context

Closes #487 

## Types of changes

## Checklist:

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ All notable changes to AET will be documented in this file.
 - [PR-489](https://github.com/Cognifide/aet/pull/489) Cleaner integration tests
 - [PR-480](https://github.com/Cognifide/aet/pull/480) Test summary stats on the main report page. ([#474](https://github.com/Cognifide/aet/issues/474))
 - [PR-459](https://github.com/Cognifide/aet/pull/459) Print more meaningful error messages when suite.xml is malformed ([#436](https://github.com/Cognifide/aet/issues/436)) 
-- [PR-517](https://github.com/Cognifide/aet/pull/517) Angular sends a URL name instead of a URL to RerunServlet ([#487](https://github.com/Cognifide/aet/issues/487))
+- [PR-517](https://github.com/Cognifide/aet/pull/517) Single URL rerun fixed for named URLs ([#487](https://github.com/Cognifide/aet/issues/487))
 
 ## Version 3.2.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to AET will be documented in this file.
 - [PR-489](https://github.com/Cognifide/aet/pull/489) Cleaner integration tests
 - [PR-480](https://github.com/Cognifide/aet/pull/480) Test summary stats on the main report page. ([#474](https://github.com/Cognifide/aet/issues/474))
 - [PR-459](https://github.com/Cognifide/aet/pull/459) Print more meaningful error messages when suite.xml is malformed ([#436](https://github.com/Cognifide/aet/issues/436)) 
+- [PR-517](https://github.com/Cognifide/aet/pull/517) Angular sends a URL name instead of a URL to RerunServlet ([#487](https://github.com/Cognifide/aet/issues/487))
 
 ## Version 3.2.2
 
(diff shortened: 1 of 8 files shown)


report/src/main/webapp/app/layout/toolbar/toolbarBottom.view.html
```diff
@@ -35,7 +35,7 @@ <h4 class="toolbar-link-header ellipsis">
         data-content="Rerun current URL"
         data-trigger="hover"
         data-container="body"
-        data-ng-click="toolbarBottom.rerunURL(toolbarBottom.model.url)">
+        data-ng-click="toolbarBottom.rerunURL(toolbarBottom.model.name)">
```
tkaik: Will it also work when URL doesn't have a "name" attribute in XML?

Cich0sza: Yes, because when user doesn't set "Name" attribute, then URL is set as "Name"

report/src/main/webapp/app/services/rerun.service.js
```diff
@@ -81,10 +86,14 @@
     function rerunURL(testName, testUrl) {
       $rootScope.rerunInProgress = true;
       var suiteInfo = suiteInfoService.getInfo();
-      var rerunParams = 'company=' + suiteInfo.company + '&' + 'project=' + suiteInfo.project + '&' +
-        'suite=' + suiteInfo.name + '&' + 'testUrl=' + testUrl + '&' + 'testName=' + testName;
-      var url = endpointConfiguration.getEndpoint().getUrl + 'suite-rerun?' + rerunParams;
-      $http.post(url, {}).then(function successCallback(response) {
+      var url = endpointConfiguration.getEndpoint().getUrl + 'suite-rerun';
+      $http.post(url, {
```
plutasnyy: It warms my heart :heart:

test-executor/src/main/java/com/cognifide/aet/executor/http/RerunDataWrapper.java
```diff
@@ -0,0 +1,102 @@
...
+public class RerunDataWrapper {
+
+    private String correlationId;
+    private String company;
+    private String project;
+    private String suite;
+    private String testUrl;
+    private String testName;
+
+    public RerunDataWrapper(String correlationId, String company, String project, String suite, String testUrl, String testName) {
```
tkaik: This constructor is probably redundant, I don't think that GSON is using it. Setters are also redudant here

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ All notable changes to AET will be documented in this file.
 - [PR-489](https://github.com/Cognifide/aet/pull/489) Cleaner integration tests
 - [PR-480](https://github.com/Cognifide/aet/pull/480) Test summary stats on the main report page. ([#474](https://github.com/Cognifide/aet/issues/474))
 - [PR-459](https://github.com/Cognifide/aet/pull/459) Print more meaningful error messages when suite.xml is malformed ([#436](https://github.com/Cognifide/aet/issues/436)) 
+- [PR-517](https://github.com/Cognifide/aet/pull/517) Angular sends a URL name instead of a URL to RerunServlet ([#487](https://github.com/Cognifide/aet/issues/487))
```
tkaik: I suggest simplifying this, e.g. `Single URL rerun fixed for named URLs` so it's more understandable :)

PR #518, opened 2019-08-08
Bugfix/resolution and subimage issues

### LINKED ISSUE #384: Collector hangs when auto height feature sets the height to 0
### Version used

* `AET 3.0.1`

### Context of the issue/feature

When you use the Resolution Modifier without the `height` parameter for a page which has `0` height (i.e. the `document.body.scrollHeight` JS command returns `0`), the collector hangs infinitely - client receives `Timeout was reached while receiving status message` error.

Expected behaviour can be one of following: 
* throw processing error/warning for screenshot of given URL with information about the page with `0` height
* when the calculated height is `0`, set it to some default low value instead (e.g. even `1px` can be enough to prevent this issue)

### Steps to reproduce

* Preapare a suite with auto-height calulcation in `collect` phase
* Run the suite on a page which has `0` height
* Client receives timeout

Cich0sza:
Fixed bug in [Resolution Modifier](https://github.com/Cognifide/aet/wiki/ResolutionModifier)

## Description

When user did not set `height` parameter, it is calculated automatically. To prevent error, minimum height is set to 1px, when site height is 0px.

## Motivation and Context

Closes #384

## Types of changes

## Checklist:

---


core/jobs/src/main/java/com/cognifide/aet/job/common/collectors/screen/ScreenCollector.java
```diff
@@ -222,6 +224,25 @@ public void setParameters(Map<String, String> params) throws ParametersException
...
+  private BufferedImage getSubImage(BufferedImage fullImg, Point point, Dimension size) {
+    int width = size.getWidth();
+    int height = size.getHeight();
+    if (point.getX() + width > fullImg.getWidth()) {
+      width = fullImg.getWidth() - point.getX();
+    }
+    if (point.getY() + height > fullImg.getHeight()) {
+      height = fullImg.getHeight() - point.getY();
+    }
+    if (point.getX() + width == 0) {
```
tkaik: I think that instead of those 2 `if`'s, we can change `if` above to:
` width = Math.max(1, fullImg.getWidth() - point.getX()` (same for height)

Cich0sza: done ;)

integration-tests/test-suite/partials/layout.xml
```diff
@@ -532,5 +532,62 @@
...
+		</collect>
+		<compare>
+			<screen comparator="layout"/>
+		</compare>
+		<urls>
+			<url href="comparators/layout/1px_height_page.jsp"/>
+		</urls>
+	</test>
+
+	<test name="S-comparator-Layout-Auto-Height-0px-Page">
```
plutasnyy: I am not 100% sure but with change in integration tests you probably should update also https://github.com/Cognifide/aet/blob/master/integration-tests/sanity-functional/src/test/resources/features/filtering.feature or/and https://github.com/Cognifide/aet/blob/master/integration-tests/sanity-functional/src/test/java/com/cognifide/aet/sanity/functional/HomePageTilesTest.java

I remember that documentation contains some tutorial how to run these tests but unfortunatelly I cannot find it :P

PR #522, opened 2019-08-27
Bugfix/improve performance of generating /configs/list response

### LINKED ISSUE #519: Performance of generating /configs/list response
**Describe the bug**
For large AET databases, getting response from /configs/list endpoint takes a very long time and consumes a lot of system resources (RAM, CPU) which might heavily impact performance of running suites in the system.

* Version used: `AET 3.2.2`

**To Reproduce**
1. Prepare AET environment with large database (e.g. at least 100 GB in size / few hundreds of suites, each with few hundred versions)
2. Go to <karaf-url>/config/list
3. Wait for reponse

**Expected behavior**
Response should be returned quickly even for large databases. Response content may be limited to last X versions of report for each suite, instead of always returning every availalbe raport URL.

**Additional context**
Currently /configs/list endpoint is building full HTML response with StringBuilder. Consider chaning this approach. It also may be worth to check if executed DB queries are using indexes.

Cich0sza:
Performance of generating /configs/list was improved.

## Description

The proposed solution uses the FreeMarker template engine to create an HTML response.
 /configs/list endpoint has now three optional parameters, `company`, `project`, and `name`. The number of parameters determines which view will be generated. 

- When a request doesn't contain parameters the project list is generated
- When the `company` and `project` parameters are provided then the suite list is generated. 
- When all three parameters are provided then the version list is generated. 

## Motivation and Context

fixes #519 

## Types of changes

## Checklist:

---


Slasheruus:
Please correct -> /configs/list

annabudziszewska:
Tested and accepted

PR #524, opened 2019-09-23
Remove handling unescaped URLs in suites

wblachowski:
Removed escaping URLs in suite XMLs. 
As a result, suite file has to be a valid XML file to be successfully processed.

## Description
This change removes additional escaping of URLs present in suites. Current behaviour expects all URLs to be unescaped and thus suite files are often invalid XMLs. Escaping URLs on your own leads to bugs described in #441. 

By enforcing escaped URLs we make sure all suite files are valid XML files and prevent confusion regarding this topic. In case where there are unescaped URLs in a suite file, there will be an according error displayed, e.g:
```
Something is wrong with your suite definition! Detected errors:
Unexpected character '<' (code 60) in attribute value
 at [row,col {unknown-source}]: [24,45]
```
which make it easy to fix all incorrect files.

## Motivation and Context
This is a second approach to a bug described in #441. 
Handling both escaped and unescaped URLs was implemented in #460, however it's been discussed that it may not be the best solution. 

## Upgrade notes (if appropriate)
Migrating to the version with this feature could require updating project's suite files to ensure that all URLs are properly escaped.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

---


plutasnyy:
Is it possible that after this change some suites won't work? If yes we could add some upgrade notes, otherwise don't care :D

test-executor/src/test/java/com/cognifide/aet/executor/xmlparser/xml/XmlTestSuiteParserTest.java
```diff
@@ -58,4 +58,14 @@ public void testParse_withWrongUrl_inTwoTestCase_expectParseException() throws E
     TestSuiteRun testSuite = testSuiteParser.parse(testSuitFile);
   }
 
+  @Test(expected = ParseException.class)
+  public void testParse_withUnescapedUrls_expectParseException() throws Exception {
+    URL resourceURL = getClass().getResource("/testSuiteWithUnescapedUrls.xml");
+    File testSuitFile = new File(resourceURL.toURI());
```
dzuma: typo testSuit**e**File

annabudziszewska:
Tested and accepted
Version: 3.3.1.SNAPSHOT

PR #412, opened 2018-10-23
Rerun suite, test, url

plutasnyy:
## Description
This PR provide you the functionality to rerun the whole suite - it runs in the same way like typical suite run, only test or url rerun - rerun part of suite work in a different way, after rerun, artifacts in your suite will be overwritten and aet doesn't create a new version of the suite.
The new way to run a test for page requires changes in the cleaner. The cleaner should remove overwritten artifacts- we don't want rubbish in our database. Functionality for the cleaner is in this PR

## Motivation and Context
It would be a nice improvement for users

## Screenshots (if appropriate):
44725131-e82f4200-aad4-11e8-9466-4d1994da8ca7
44725134-e9f90580-aad4-11e8-9b1a-59263d1b76ff
46950715-ec63ee80-d085-11e8-98db-9d7c609afb83
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the [code style](https://github.com/Cognifide/aet/blob/master/CONTRIBUTING.md#coding-conventions) of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

---
I hereby agree to the terms of the AET Contributor License Agreement.

PR #526, opened 2019-10-02
Add sending urls in packets

### LINKED ISSUE #431: AET sends always one url in package
With reference to last new feature in AET's - rerun, I am obligated to describe a bit about changes in CollectDispatcher class. Some time ago, every open a new browser needed a lot of resources. It was the reason why URLs were compressing into packages kept by MessageWithDestination and we used every browser to test a whole package of URLs. During the implementation of the rerun, we had to do huge refactor of Runner and we decided to leave it functionality and replaced it by sending always one message.

wblachowski:
This change reintroduces sending multiple URLs in one message to collectors in order to boost performance.

## Description
Changes made as a part of rerun functionality in #412 (in `CollectDispatcher` class among others) removed the possibility to send multiple URLs in a single message received by `CollectorMessageListener`. This affected the performance because there is a new web driver created for each received message and this operation is resource consuming. By grouping multiple URLs in a single message it is possible to reuse a single web driver many times.

I made some measurements using a dummy suit with 2 tests, each containing 10 URLs pointing to a server running on my local machine. The average times of 3 test runs for different URL package sizes are present in the following table:

| Package size  | Average suite runtime (in seconds) |
| ------------- | ------------- |
| 1  | 33.22  |
| 3  | 24.07  |
| 6  | 19.22  |
| 9  | 27.04  |

Please keep in mind that only URLs within a single test are grouped (because each test can have different proxy setting etc.) so the performance gain is heavily dependent on the suit being tested,

## Motivation and Context
Closes #431 

## Types of changes

## Checklist:

---


annabudziszewska:
Tested and accepted
Version: 3.3.1.SNAPSHOT

Issue #500, opened 2019-04-02
Add source pattern date and source collected date


RdkLwg:
As the user I would like to see the date when the source was set as the pattern and collected last time (the same as it is for visual comparison) in the AET Report.

PR #528, opened 2019-10-14
add source pattern and collected date

KarolFrankiewicz:
This change adds dates for pattern and collected source in source comparator.

## Description
User is able to see dates of collected and pattern sources just like it works right now for screenshots comparison
In case pattern date was not saved before, both dates won't be shown

## Motivation and Context
feature required in #500 

## Types of changes

## Checklist:

---


core/jobs/src/main/java/com/cognifide/aet/job/common/comparators/source/SourceComparator.java
```diff
@@ -119,4 +120,16 @@ private void addFormattedSources(ComparatorStepResult result) {
     result.addData("formattedSource", data);
     result.addData("sourceCompareType", sourceCompareType.name());
   }
+
+  private void addTimestampToResult(ComparatorStepResult result) {
+    long stamp;
+    try {
+      stamp = artifactsDAO.getArtifactUploadDate(properties, properties.getPatternId()).getTime();
+    } catch(Exception e) {
+      stamp = 0;
```
dzuma: never loose exception log :)

core/jobs/src/main/java/com/cognifide/aet/job/common/comparators/source/SourceComparator.java
```diff
@@ -119,4 +120,16 @@ private void addFormattedSources(ComparatorStepResult result) {
...
+    }
+    result.addData("patternTimestamp", Long.toString(stamp));
```
tMaxx, line 131: This should be done a bit differently: instead of doing it inside try-catch and zeroing `stamp`, you should put the `result.addData("patternTimestamp", ...)` inside `try` block (with properly labelled noop/debug log for `catch`), and then conditionally display it inside the view of report.

report/src/main/webapp/app/layout/main/url/reports/source-diff.html
```diff
@@ -19,11 +19,15 @@
 -->
 <div class="clearfix"></div>
 <div class="col-xs-6">
+  <p class="layout-compare-description text-center">Pattern
+    <span class="label label-info">{{case.comparator.stepResult.data.patternTimestamp | date:'yyyy-MM-dd HH:mm'}}</span><br>
+    <span class="label-spacer"></span>
+  </p>
```
tMaxx, line 25: See the comment for `SourceComparator`. The timestamp should ideally be displayed only when it's available and truthy.

KarolFrankiewicz: after update it is shown only when both dates are available - otherwise date string is empty - hence it is not showed

core/jobs/src/main/java/com/cognifide/aet/job/common/comparators/source/SourceComparator.java
KarolFrankiewicz: done

core/jobs/src/main/java/com/cognifide/aet/job/common/comparators/source/SourceComparator.java
KarolFrankiewicz: done

tMaxx (approved):

PR #531, opened 2019-10-28
Enable using login modifier with only partial path

Skejven:
## Description
Currently, the login modifier requires passing the full `login-page` URL in the properties.

## Motivation and Context
Passing full URL to the login page is sometimes difficult, especially when a single suite is reused among multiple domains.
This PR enables running `login` modifier with only part of the login page, and the full URL is built using `domain` (the same logic as `url` works).

## Upgrade notes (if appropriate)
This feature is backward compatible. Proper updates were made in the wiki docs.
The logic behind the change is:
- if the `login-page` starts with `http` prefix, then the old logic is triggered (full login page URL)
- otherwise, `login-page` is treated as a path, and `domain` is added in the front (see unit tests for details).

## Types of changes

## Checklist:

---


annabudziszewska:
Tested and accepted

PR #547, opened 2020-03-16
New logo assets plus necessary path changes

RdkLwg:
New logo assets and necessary changes in paths/file names


RdkLwg:
Thanks for nice hints. The first item has been already addressed with additional variant of the logo.
The other one with the second commit.

